PR #7, opened 2016-03-13
Check in schema.rb and configure Travis to run tests.

jfhamlin:
This gets Travis builds succeeding. Not too hard, since we don't have any tests yet. A few things to note:

1. We haven't yet installed rspec yet, so the test command will change when we move to it.
2. We should eventually ditch .rubocop_todo.yml, excluding rails/bundle files in .rubocop.yml and fixing any violations in our app code.


jfhamlin:
cc @lelandg559 @jjfreund

.rubocop.yml
```diff
@@ -1,8 +1 @@
-AllCops:
```
jfhamlin: We should probably keep this stuff.

jfhamlin: This key is deprecated in the latest versions of rubocop.

lelandg559:
LGTM

PR #12, opened 2016-03-17
Remove unused gems, install rspec and initialize spec/ dir.

jfhamlin:
We'll be using rspec rather than Rails' default testing framework. This change installs rspec and removes many gems we aren't using (and won't use, since we're not serving HTML/CSS). Rails commands like `rails generate model MyModel` will also generate boilerplate rspec files. An example:

```
$ bundle exec rails generate model Foo
      invoke  active_record
      create    db/migrate/20160317031915_create_foos.rb
      create    app/models/foo.rb
      invoke    rspec
      create      spec/models/foo_spec.rb
```


jfhamlin:
Reviewed with @jjfreund

PR #14, opened 2016-03-20
HTTP endpoints for reading resources,categories, and resources by category resolves #4

### LINKED ISSUE #4: Create API access to phase one domain

lelandg559:


lelandg559:
The paths for now don't require any sort of session or auth, and are just available at

/api/resource
/api/category
/api/resource?categoryId={id}

Gemfile
```diff
@@ -3,6 +3,7 @@ source 'https://rubygems.org'
 gem 'rails', '4.2.6'
 gem 'pg', '~> 0.15'
 gem 'sass-rails', '~> 5.0'
+gem 'rails-api'
```
jfhamlin: :+1: We should probably lock this down, especially since it's pre-1.0. `gem 'rails-api', '0.4.0' should do it.

Gemfile
```diff
@@ -3,6 +3,7 @@ source 'https://rubygems.org'
 gem 'rails', '4.2.6'
 gem 'pg', '~> 0.15'
 gem 'sass-rails', '~> 5.0'
```
jfhamlin: Looks like I didn't get rid of all the unnecessary dependencies. I'll pare this down further once your changes are in master.

app/controllers/api/category_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Api::CategoryController < ApplicationController
+  def index
+    @categories = Category.all
```
jfhamlin: Since the field isn't doing any work here, this could probably just be `render json: Category.all`.

config/routes.rb
```diff
@@ -53,4 +53,9 @@
   #     # (app/controllers/admin/products_controller.rb)
   #     resources :products
   #   end
+
+  namespace :api do
```
jfhamlin: I think we can omit the 'api' namespace.

config/routes.rb
```diff
@@ -53,4 +53,9 @@
...
+    resources :category
```
jfhamlin: The convention is to use the plural name of the collection. See some of the examples in the comments in this file, above. The controllers should be renamed to use the plural as well, e.g. `CategoriesController`.

app/controllers/application_controller.rb
```diff
@@ -1,4 +1,5 @@
-class ApplicationController < ActionController::Base
+class ApplicationController < ActionController::API
+  include ActionController::RequestForgeryProtection
```
jfhamlin: I think we can ditch the request forgery protection altogether (including the method call below). I don't think it'll help with our use case.

app/controllers/resources_controller.rb
```diff
@@ -0,0 +1,12 @@
+class ResourcesController < ApplicationController
+  def index
+    category_id = params[:category_id]
+    @resources = if category_id
+                   Category.find(category_id).resources
```
jfhamlin: An alternative:

 `Resource.where category_id: category_id`

There are semantic and performance differences between the two. `Category.find ...` will result in a 404 if the category doesn't exist, whereas `Resource.where ...` will simply return no results. `Category.find ...` performs two BD queries, one for the category and a second for the resources; whereas `Resource.where ...` performs one.

lelandg559: got it, yeap, not familiar with the syntax to do it optimally

jfhamlin: Yeah, no problem. There's a learning curve to `ActiveRecord`. It's not immediately clear what it ends up doing under the hood. It takes experience.

jfhamlin: For posterity: @lelandg559 reminded me that we've got a join table between the two, so what I've got doesn't work. We have to get a bit fancier to make it work.

jfhamlin:
LGTM once Travis says it's :green_apple:.

PR #15, opened 2016-03-21
Pare down Gemfile further and remove initial comments.

jfhamlin:
This change removes the 'rails' gem from the Gemfile, replacing it with the pieces of rails we actually need. Instructional comments included in initial files are also removed.


Gemfile
```diff
@@ -1,10 +1,15 @@
 source 'https://rubygems.org'
 
-gem 'rails', '4.2.6'
```
lelandg559: so rails doesn't need rails?

jfhamlin: This change pulls in the specific rails components we need below ('actionmailer', 'activerecord', etc.), leaving out components for things like view rendering which would be included by the full rails gem. It's not vital that we remove these dependencies, but I kinda like being a bit more minimal.

lelandg559:
LGTM , ran locally

PR #17, opened 2016-03-21
[migration] Add unique index to categories on name.

jfhamlin:
Two categories shouldn't be able to share a name. This change adds an index on the column to enforce that constraint.

We should discuss the API in a bit more detail at some point, too. It might make more sense for clients to use the category name as a category identifier rather than its `id` field.

### DIFF
--- a/db/migrate/20160321021009_add_index_to_categories_name.rb
+++ b/db/migrate/20160321021009_add_index_to_categories_name.rb
@@ -1,0 +1,5 @@
+class AddIndexToCategoriesName < ActiveRecord::Migration
+  def change
+    add_index :categories, :name, unique: true
+  end
+end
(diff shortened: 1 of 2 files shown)


lelandg559:
leland@ophion:~/askdarcel-api$ rake db:migrate
== 20160321021009 AddIndexToCategoriesName: migrating =========================
-- add_index(:categories, :name, {:unique=>true})
   -> 0.0169s
== 20160321021009 AddIndexToCategoriesName: migrated (0.0170s) ================

leland@ophion:~/askdarcel-api$ rails server
=> Booting WEBrick
=> Rails 4.2.6 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
=> Ctrl-C to shutdown server
[2016-03-21 16:13:22] INFO  WEBrick 1.3.1
[2016-03-21 16:13:22] INFO  ruby 2.2.2 (2015-04-13) [x86_64-darwin15]
[2016-03-21 16:13:22] INFO  WEBrick::HTTPServer#start: pid=78440 port=3000

LGTM

PR #16, opened 2016-03-21
Add resource, category factories and resources request specs.

jfhamlin:
- Adds factory_girl_rails gem.
- Adds two simple request specs to test the basic resource API.

PR #18, opened 2016-03-22
Add a rake task to populate a development database.

jfhamlin:
This includes the commit from #16. We'll want to wrap that PR up and get it in before merging this one.

### DIFF
--- a/lib/tasks/populate.rake
+++ b/lib/tasks/populate.rake
@@ -1,0 +1,29 @@
+namespace :db do
+  desc 'Erase and fill a development database'
+  task populate: :environment do
+    unless Rails.env.development?
+      puts 'db:populate task can only be run in development.'
+      exit
+    end
+
+    [Category, Resource].each(&:delete_all)
+
+    FactoryGirl.create_list :category, 20
+
+    categories = Category.all
+
+    128.times do
+      name = Faker::Company.name
+      short_description = Faker::Lorem.sentence if rand(2) == 0
+      long_description = Faker::Lorem.paragraph if rand(2) == 0
+      website = Faker::Internet.url if rand(2) == 0
+
+      FactoryGirl.create(:resource,
+                         name: name,
+                         short_description: short_description,
+                         long_description: long_description,
+                         website: website,
+                         categories: categories.sample(rand(4)))
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


lib/tasks/populate.rake
```diff
@@ -0,0 +1,35 @@
...
+      FactoryGirl.create(:category,
+                         name: Faker::Lorem.sentence(rand(4))
+                        )
+    end
+
+    categories = Category.all
+
+    128.times do
+      name = Faker::Company.name
+      short_description = Faker::Lorem.sentence if rand(2) == 0
```
jfhamlin: We could also just add logic like this to the base factory.

lelandg559:
Ocassionally seeing this because of random duplicate on unique index name

leland@ophion:~/askdarcel-api$ rake db:populate
rake aborted!
ActiveRecord::RecordNotUnique: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_categories_on_name"
DETAIL:  Key (name)=(.) already exists.
: INSERT INTO "categories" ("name", "created_at", "updated_at") VALUES ($1, $2, $3) RETURNING "id"
/Users/leland/askdarcel-api/lib/tasks/populate.rake:12:in `block (3 levels) in <top (required)>'
/Users/leland/askdarcel-api/lib/tasks/populate.rake:11:in `times'
/Users/leland/askdarcel-api/lib/tasks/populate.rake:11:in `block (2 levels) in <top (required)>'
PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_categories_on_name"
DETAIL:  Key (name)=(.) already exists.
/Users/leland/askdarcel-api/lib/tasks/populate.rake:12:in `block (3 levels) in <top (required)>'
/Users/leland/askdarcel-api/lib/tasks/populate.rake:11:in `times'
/Users/leland/askdarcel-api/lib/tasks/populate.rake:11:in `block (2 levels) in <top (required)>'
Tasks: TOP => db:populate
(See full trace by running task with --trace)




Other issue is that the join table isn't being appropriately filled up.

jfhamlin:
> Ocassionally seeing this because of random duplicate on unique index name

Yeah, I originally just let that be, but it'll get annoying, so I've fixed that.

> Other issue is that the join table isn't being appropriately filled up.

Good catch. Fixed that too.

lelandg559:
yeap LGTM

PR #19, opened 2016-03-22
Allow cross origin requests for client ajax calls.

schroerbrian:
allows ajax requests from client web app.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,2 +1,9 @@
 class ApplicationController < ActionController::API
+  skip_before_filter :verify_authenticity_token
+  before_filter :add_allow_credentials_headers
+
+  def add_allow_credentials_headers
+    response.headers['Access-Control-Allow-Origin'] = request.headers['Origin'] || '*'
+    response.headers['Access-Control-Allow-Credentials'] = 'true'
+  end
 end


app/controllers/application_controller.rb
```diff
@@ -1,2 +1,10 @@
 class ApplicationController < ActionController::API
+  skip_before_filter :verify_authenticity_token
+  before_filter :add_allow_credentials_headers
+
+  def add_allow_credentials_headers
```
lelandg559: any difference in how this would be configured in dev vs prod env?

app/controllers/application_controller.rb
```diff
@@ -1,2 +1,10 @@
 class ApplicationController < ActionController::API
+  skip_before_filter :verify_authenticity_token
```
jfhamlin: Hmm. If this filter is still enabled, I wonder if we can find somewhere in the config to disable it and any default filters like it. I don't think we'll need them for an API-only rails server.

app/controllers/application_controller.rb
```diff
@@ -1,2 +1,10 @@
...
+  before_filter :add_allow_credentials_headers
+
+  def add_allow_credentials_headers                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
+    response.headers['Access-Control-Allow-Origin'] = request.headers['Origin'] || '*'                                                                                                                                                                                                     
+    response.headers['Access-Control-Allow-Credentials'] = 'true'
```
jfhamlin: Reading up on this, it seems it's used to indicate to the browser that it's OK to send cookies. I don't think we should be sending cookies with requests if we can help it.

schroerbrian: ah, i hadn't know that this configures cookies as well. could you point me to the resource that discusses this? Perhaps he best practice for enabling CORS on rails app is through rack middleware? http://stackoverflow.com/a/29751501

lelandg559:
Works for me, going to merge this into master so that UI/API can talk.

PR #26, opened 2016-04-04
Jjfreund/sf link clone domain

jjfreund:
A minor refactor awaits. I would prefer to combine it with another pending refactor, and what we have should be sufficient for milestone 2.


lelandg559:
LGTM

PR #27, opened 2016-04-05
Added nested JSON objects for schedule/addresses/phones on resource Added find resource by id

lelandg559:
Requires you to drop database and rebuild

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -1,12 +1,17 @@
 class ResourcesController < ApplicationController
   def index
     category_id = params[:category_id]
-    @resources = if category_id
-                   Resource.joins(:categories).where('categories.id' => category_id)
-                 else
-                   Resource.all
-                 end
+    resources = if category_id
+                  Resource.joins(:categories).where('categories.id' => category_id)
+                else
+                  Resource.all
+                end
 
-    render json: @resources
+    render json: resources.to_json(include: ['addresses', 'phones', { schedule: { include: 'schedule_days' } }])
+  end
+
+  def show
+    resource = Resource.find(params[:id])
+    render json: resource.to_json(include: ['addresses', 'phones', { schedule: { include: 'schedule_days' } }])
   end
 end
--- a/spec/requests/resources_spec.rb
+++ b/spec/requests/resources_spec.rb
@@ -32,4 +32,18 @@ RSpec.describe 'Resources' do
       end
     end
   end
+  context 'show' do
+    let!(:resources) { create_list :resource, 4 }
+    let!(:resource_a) { create :resource, name: 'a' }
+
+    it 'returns specific resource' do
+      get "/resources/#{resource_a.id}"
+      expect(response_json).to include(
+        'id' => resource_a.id,
+        'addresses' => Array,
+        'schedule' => Hash,
+        'phones' => Array
+      )
+    end
+  end
 end
--- a/app/models/resource.rb
+++ b/app/models/resource.rb
@@ -1,4 +1,6 @@
 class Resource < ActiveRecord::Base
   has_and_belongs_to_many :categories
   has_many :addresses
+  has_many :phones
+  has_one :schedule
 end
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -1,3 +1,4 @@
 class Schedule < ActiveRecord::Base
   belongs_to :resource
+  has_many :schedule_days
 end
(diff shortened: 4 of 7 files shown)


db/migrate/20160401043112_create_phones.rb
```diff
@@ -5,7 +5,7 @@ def change
 
       t.string :number, null: false
       t.string :extension
-      t.string :type, null: false
+      t.string :service_type, null: false
```
jfhamlin: It's usually bad practice to edit migrations for the reason you mention: devs will have to recreate their local DBs. Another migration that renames the column is a better way to go. There's a `rename_column` method for migrations that makes this pretty easy.

lelandg559: Yea agree, ill do that now.

jfhamlin: Why the rename, BTW?

app/controllers/resources_controller.rb
```diff
@@ -7,6 +7,11 @@ def index
                    Resource.all
                  end
 
-    render json: @resources
+    render json: @resources.to_json(include: ['addresses', 'phones', { schedule: { include: 'schedule_days' } }])
+  end
+
+  def show
+    @resources = Resource.find(params[:id])
```
jfhamlin: I think this can be `resource` (singular, no "@"). The `@resources` variable in `index` should probably be `resources`, too. The variables aren't referenced outside of the methods in which they're defined, so they don't need to be instance variables.

jfhamlin:
Worth a request test in https://github.com/ShelterTechSF/askdarcel-api/blob/master/spec/requests/resources_spec.rb?

lelandg559:
Added a simple test, lmk if thats what you were looking for

db/migrate/20160401043112_create_phones.rb
lelandg559: 'type' as a column name is something rails does not like.

If i can find the error I got ill post it

lelandg559: It was similiar to this

The single-table inheritance mechanism failed to locate the subclass: 'phones'. This error is raised because the column 'type' is reserved for storing the class in case of inheritance. Please rename this column if you didn't intend it to be used for storing the inheritance class or overwrite Project.inheritance_column to use another column for that information.

spec/requests/resources_spec.rb
```diff
@@ -32,4 +32,14 @@
...
     end
   end
+  context 'show' do
+    let!(:resources) { create_list :resource, 4 }
+    let!(:resource_a) { create :resource, name: 'a' }
+
+    it 'returns specific resource' do
+      get "/resources/#{resource_a.id}"
+      returned_ids = response_json['id']
+      expect(returned_ids).to match(resource_a.id)
```
jfhamlin: Was thinking we could make sure that the child resources are rendered as well. Could be:

```ruby
expect(response_json).to match(
  'id' => resource_a.id,
  'addresses' => Array,
  'schedule' => Hash,
  'phones' => Array
)
```

You could make the assertion stronger, but that at least captures the presence of nested resources.

db/migrate/20160401043112_create_phones.rb
jfhamlin: Gotcha. If we prefer the "type" column name, we can go with the second suggestion:

>  overwrite <Model>.inheritance_column to use another column for that information.

`self.inheritance_column = '_unused'` or something like that in the `Phone` model should work around this without having to change the schema.

lelandg559: Well the migration works fine so would we want to handle this in that way at this point?

Just to avoid "custom" configurations, I don't think there was any specific desire to have it called specifically "type"

jfhamlin: I slightly prefer "type" to "service_type," mainly because it's not clear what "service" refers to. But I don't have that strong a preference. Rename is fine.

spec/requests/resources_spec.rb
lelandg559: is that syntax correct? not familiar with rspec matchers but I can't find out how that would match against the reference just containing an Array of anything

jfhamlin: Ah, it should be `include`, not `match`.

jfhamlin:
@lelandg559, made the test change I mentioned.

@jjfreund, this LGTM once CI goes through.

PR #30, opened 2016-04-07
Added commands to set up local database and current ruby version

lelandg559:


lelandg559:
Just want to start convo on what we should be adding, feel free to suggest more things I can add.

README.rdoc
```diff
@@ -7,12 +7,24 @@ Things you may want to cover:
...
+2.2.2
+
 * System dependencies
 
 * Configuration
 
 * Database creation
 
+This will create the database in postgres if does not exist.
+rake db:create
```
jfhamlin: It'd be good to note that a locally-running postgres server is required.

lelandg559:
Just documenting some commands, so merging this.

PR #11, opened 2016-03-17
Docker for local dev

shannonrdunn:

PR #31, opened 2016-04-07
Revert "Merge pull request #11 from ShelterTechSF/docker"

jfhamlin:
This reverts commit 10611f65f9cc400dfd1caec5f477c03378ed4ba8, reversing
changes made to d0cc0b5759bcb2b465e764605f915a8fb68582ea.

We may want to use docker in development later, but it's currently breaking
OSX dev environments.


jfhamlin:
@shannonrdunn, I propose backing out the docker changes for now. Dev environments don't work out-of-the-box with these changes and it's not clear to newcomers (or me at this point) how to get set up with docker. Let's bring this stuff back once it's more fleshed out.

jfhamlin:
The build is broken from a previous change.

shannonrdunn:
Sounds fine.

Its just the hostname its looking for is db, and postgres user. Which are standard in the postgres docker image, locally you would just need to map db hostname to localhost, and use the postgres user instead of root.

But you're right, lets revisit once I can get a doc out explaining better.

PR #33, opened 2016-04-07
Clean up previous change to add sample resource descriptions.

jfhamlin:
Creates a faker class for random ShelterTech-specific data.


lelandg559:
LGTM

Issue #41, opened 2016-04-20
Add API end point for user registration


jjfreund:

lelandg559:
Duplicate of #53

Issue #34, opened 2016-04-20
Add auth functionality to API


jjfreund:
Users should be able to view resources and categories without logging in. Logging in should be necessary to write reviews. User and Admin roles should be supported; Admin functionality will be decided upon later.

lelandg559:
Duplicate of #53

PR #42, opened 2016-04-21
Add auth functionality to API #34

lelandg559:
Add API end point for user registration #41


app/controllers/application_controller.rb
```diff
@@ -2,6 +2,10 @@ class ApplicationController < ActionController::API
   skip_before_filter :verify_authenticity_token
   before_filter :add_allow_credentials_headers
 
+  include ActionController::Helpers
```
lelandg559: These were all required/suggested to be added by researched I did with making devise behave properly, lmk if you have any deeper insight

app/models/user.rb
```diff
@@ -0,0 +1,12 @@
+class User < ActiveRecord::Base
+  # Include default devise modules.
+  devise :database_authenticatable, :registerable,
+          :recoverable, :rememberable, :trackable, :validatable,
+          :confirmable, :omniauthable
+  include DeviseTokenAuth::Concerns::User
+  DeviseTokenAuth::Concerns::User
+  # Include default devise modules. Others available are:
+  # :confirmable, :lockable, :timeoutable and :omniauthable
+  devise :database_authenticatable, :registerable,
```
lelandg559: this class looks like it got messed up (duplication), ill fix

jfhamlin:
I wanted to test this locally to see what the requests and responses look like, but the migrations don't run successfully. It looks like both devise and devise-token-auth are trying to create a `users` table:

```
$ be rake db:migrate:status

database: askdarcel_development

 Status   Migration ID    Migration Name
--------------------------------------------------
   up     20160320194732  Create categories
   up     20160320195323  Create resources
   up     20160320200142  Create categories resources
   up     20160321021009  Add index to categories name
   up     20160401042735  Create addresses
   up     20160401043112  Create phones
   up     20160401043139  Create schedules
   up     20160401043947  Create schedule days
   up     20160405025104  Change column name phone type
   up     20160420083052  Devise create users
  down    20160420102224  Devise token auth create users
```

lelandg559:
Ok, I'll review and fix that.

PR #44, opened 2016-05-01
Jjfreund/milestone3 domain

lelandg559:


app/controllers/resources_controller.rb
```diff
@@ -6,12 +6,15 @@ def index
                 else
                   Resource.all
                 end
-
-    render json: resources.to_json(include: ['addresses', 'phones', 'categories', { schedule: { include: 'schedule_days' } }])
+    # rubocop:disable Metrics/LineLength
```
jfhamlin: How about splitting this onto multiple lines instead? It'd be easier to read that way. We could also move the inclusion hash to a constant. For example:

```ruby
RENDERED_SCHEDULE_ATTRIBUTES = {
  include: 'schedule_days' 
}

RENDERED_SERVICE_ATTRIBUTES = {
  include: ['notes', { schedule: RENDERED_SCHEDULE_ATTRIBUTES }]
}

RENDERED_RESOURCE_ATTRIBUTES = {
  include: ['addresses', 'phones', 'categories', 'notes', {
    services: RENDERED_SERVICE_ATTRIBUTES
   }, {
    schedule: RENDERED_SCHEDULE_ATTRIBUTES
  }]
}
```

Then: `render json: resources.to_json(RENDERED_RESOURCE_ATTRIBUTES)`.

Poor man's presenters.

lelandg559: sure, hadn't finished up presenter commit

PR #45, opened 2016-05-02
Added services to domain model

lelandg559:
Messed up the previous pull request when I was squashing


lelandg559:
and i broke some tests. will fix

app/controllers/resources_controller.rb
```diff
@@ -6,12 +6,27 @@ def index
...
 
   def show
     resource = Resource.find(params[:id])
-    render json: resource.to_json(include: ['addresses', 'phones', 'categories', { schedule: { include: 'schedule_days' } }])
+    render json: resource.to_json(RENDERED_RESOURCE_ATTRIBUTES)
   end
+
+  RENDERED_SCHEDULE_ATTRIBUTES = {
+    include: 'schedule_days'
+  }.freeze
```
lelandg559: so rubocop wants me to add the .freeze because it thinks that these should be immutable, but its causing 

Completed 500 Internal Server Error in 147ms (ActiveRecord: 11.4ms)

RuntimeError (can't modify frozen Hash):
  app/controllers/resources_controller.rb:14:in `show'

When invoked, Ill have to have it ignore the freeze/mutability check.

app/controllers/resources_controller.rb
```diff
@@ -6,12 +6,29 @@ def index
...
+    render json: resources.to_json(RENDERED_RESOURCE_ATTRIBUTES)
   end
 
   def show
     resource = Resource.find(params[:id])
-    render json: resource.to_json(include: ['addresses', 'phones', 'categories', { schedule: { include: 'schedule_days' } }])
+    render json: resource.to_json(RENDERED_RESOURCE_ATTRIBUTES)
   end
+
+  # rubocop:disable MutableConstant
```
lelandg559: so rubocop wants me to add the .freeze because it thinks that these should be immutable, but its causing

Completed 500 Internal Server Error in 147ms (ActiveRecord: 11.4ms)

RuntimeError (can't modify frozen Hash):
app/controllers/resources_controller.rb:14:in `show'

When invoked, Ill have to have it ignore the freeze/mutability check.

jfhamlin: Ah, thanks for catching this. I haven't looked at what `to_json` does under the hood to mutate the hash, but it could conceivably alter it in such a way that subsequent calls behave in unexpected ways. I'm making some changes now to improve the performance of the resource index endpoint; while I'm at it I'll move this stuff to a method instead so we get a fresh hash every time. Seems like bad practice for `to_json` to mutate the argument though :-/.

app/controllers/resources_controller.rb
```diff
@@ -1,17 +1,36 @@
...
-    resource = Resource.find(params[:id])
-    render json: resource.to_json(include: ['addresses', 'phones', 'categories', { schedule: { include: 'schedule_days' } }])
+    resource = resources.find(params[:id])
+    render json: resource.to_json(resource_inclusion)
+  end
+
+  private
+
+  def resources
+    Resource.includes(:addresses, :phones, :categories,
```
jfhamlin: @lelandg559 This should speed up the `index` action. Without this, we were making queries for each of the associated records for every rendered resource.

lelandg559: cool, I'll try it out

app/controllers/resources_controller.rb
```diff
@@ -1,17 +1,36 @@
...
+                      schedule: :schedule_days)
+  end
+
+  def resource_inclusion
+    schedule_attrs = { include: :schedule_days }
+    service_attrs = [:notes, schedule: schedule_attrs]
```
lelandg559: So this was actually causing a 
no implicit conversion of Symbol into Integer
app/controllers/resources_controller.rb:10:in `index'

If I wrap this array in a Hash it works though, is that the right fix?
    service_attrs = { include: [:notes, schedule: schedule_attrs]}

jfhamlin: Oh, yeah. Good catch! Bad copy-paste there; this should have been wrapped in a hash as you're doing. Looks like we need a test for the new nested models, since the specs we've got didn't catch it.

jfhamlin:
@lelandg559, updated to fix the problem you uncovered and to add a test for service rendering.

lelandg559:
Are you good to put this in master at this point?

PR #53, opened 2016-05-05
Add auth functionality to API #34

lelandg559:
Add API end point for user registration #41


lelandg559:
Fix user table issue

jfhamlin:
What do you think about adding some request specs that demonstrate how a client would use the auth API?

lelandg559:
Sure, I can basically just link the devise documentation but I'll add a short example to our documentation.

jfhamlin:
> Sure, I can basically just link the devise documentation but I'll add a short example to our documentation.

Examples in the form of tests would be good too, I think. That way we can be sure we've configured Devise properly and that the API really behaves as we expect.

I don't think there's much urgency to get this merged, as authentication isn't part of the next milestone, and AFAIK we haven't solidified authentication requirements yet. Well, this is part of milestone 4 in Github, but should it be?

lelandg559:
The only near term thing this contains is the data model for Users which joe wants to use to tie in ratings/reviews.

PR #59, opened 2016-05-19
Add presenters for resources API call

lelandg559:


app/presenters/addresses_presenter.rb
```diff
@@ -0,0 +1,11 @@
...
+  property :attention
+  property :address_1
+  property :address_2
+  property :address_3
+  property :address_4
+  property :city
+  property :state_province
+  property :postal_code
+  property :country
+end
```
jfhamlin: Let's get some new lines at the end of these files. Your editor can probably be configured to add them on save.

lelandg559: You are almost faster than Rubocop

jfhamlin: Lol

app/controllers/resources_controller.rb
```diff
@@ -2,35 +2,16 @@ class ResourcesController < ApplicationController
   def index
     category_id = params[:category_id]
     result = if category_id
-               resources.joins(:categories).where('categories.id' => category_id)
```
jfhamlin: The `resources` method is doing the prefetching to avoid O(n) queries (but as I mentioned tonight, it's not prefetching enough). We should probably add it back, and I'll what's missing in my next PR.

lelandg559:
Ok, let me slow down here and push something that works, coming soon.

app/controllers/resources_controller.rb
lelandg559: oh got it

lelandg559:
@achadha this does change the response body structure slightly, the result JSON is ether wrapped in "resource" or "resources" JSON key depending on whether its looked up by ID or done by category/all.

spec/requests/resources_spec.rb
```diff
@@ -7,8 +7,8 @@
 
       it 'returns all resources' do
         get '/resources'
-        returned_ids = response_json.map { |r| r['id'] }
-        expect(returned_ids).to match(resources.map(&:id))
+        returned_names = response_json['resources'].map { |r| r['name'] }
```
jfhamlin: I think we want to leave IDs on resources for now. Otherwise clients won't be able to identify particular resources in URLs.

lelandg559: ewps, yes

jfhamlin:
LGTM

lelandg559:
I will merge this along with require UI change at same time, do not merge.

PR #81, opened 2016-06-12
Adding User migrations

lelandg559:


db/migrate/20160420102224_devise_token_auth_create_users.rb
```diff
@@ -0,0 +1,54 @@
+class DeviseTokenAuthCreateUsers < ActiveRecord::Migration
+  def change
+    create_table(:users) do |t|
+      ## Required
+      t.string :provider, :null => false, :default => "email"
+      t.string :uid, :null => false, :default => ""
+
+      ## Database authenticatable
+      t.string :encrypted_password, :null => false, :default => ""
```
jfhamlin: We aren't starting with password auth, are we?

db/migrate/20160420102224_devise_token_auth_create_users.rb
```diff
@@ -0,0 +1,54 @@
...
+      t.string   :last_sign_in_ip
+
+      ## Confirmable
+      t.string   :confirmation_token
+      t.datetime :confirmed_at
+      t.datetime :confirmation_sent_at
+      t.string   :unconfirmed_email # Only if using reconfirmable
+
+      ## Lockable
+      # t.integer  :failed_attempts, :default => 0, :null => false # Only if lock strategy is :failed_attempts
```
jfhamlin: Let's remove commented-out code.

db/migrate/20160420102224_devise_token_auth_create_users.rb
```diff
@@ -0,0 +1,54 @@
...
+      # t.string   :unlock_token # Only if unlock strategy is :email or :both
+      # t.datetime :locked_at
+
+      ## User Info
+      t.string :name
+      t.string :nickname
```
jfhamlin: Can we pare this down to what we know we'll use?

lelandg559: The thought was that if we were going to use an auth framework we would want to use what it generates and not build our own model, if it sounds like we are heading towards wanting this functionality in the medium future would want trim the domain and add it back slowly? 

If you think we should head that way ill just get the exact items that Joe needs to use, (I think it is just a user ID)

jfhamlin: I expect we can use bits and pieces of Devise, and if we need more, we can add the columns later. Maybe I'm wrong about that. But yeah, if we don't know how devise is going to solve our particular auth problems, we can start with a bare users table.

PR #84, opened 2016-06-16
Set get all resources to return HTTP 400

lelandg559:


spec/requests/resources_spec.rb
```diff
@@ -7,8 +7,7 @@
 
       it 'returns all resources' do
         get '/resources'
-        returned_ids = response_json['resources'].map { |r| r['id'] }
-        expect(returned_ids).to match(resources.map(&:id))
+        expect(response_json).to match({})
```
jfhamlin: Oh, another expectation would be nice here:

```ruby
expect(response).to have_http_status(: bad_request)
```

Should have been expecting a 200 here before.

lelandg559: sure, ill add

jfhamlin:
LGTM

PR #86, opened 2016-06-23
 Initial version of server side resource location sorting

lelandg559:


app/controllers/resources_controller.rb
```diff
@@ -1,14 +1,22 @@
...
     if result
+      if latitude and longitude
+        result = result.sort_by{ |resource| resource.address.distance_to([latitude,longitude])}
+      else
+        result = result.sort_by{ |resource| resource.__id__}
+      end
       render json: ResourcesPresenter.present(result)
     else
-      render json: {}, status: :bad_request
+      render json: {}, status: :bad_requests
```
lelandg559: thats a mistake (typo)

app/controllers/resources_controller.rb
```diff
@@ -1,14 +1,22 @@
...
+
     result = if category_id
                resources.joins(:categories).where('categories.id' => category_id)
              end
 
     if result
+      if latitude and longitude
+        result = result.sort_by{ |resource| resource.address.distance_to([latitude,longitude])}
+      else
+        result = result.sort_by{ |resource| resource.__id__}
```
lelandg559: dont need to sort here, can remove

app/presenters/resources_presenter.rb
```diff
@@ -7,7 +7,7 @@ class ResourcesPresenter < Jsonite
   property :services, with: ServicesPresenter
   property :schedule, with: SchedulesPresenter
   property :phones, with: PhonesPresenter
-  property :addresses, with: AddressesPresenter
+  property :address, with: AddressesPresenter
```
jfhamlin: Should we rename `AddressesPresenter` to `AddressPresenter`?

Gemfile
```diff
@@ -7,6 +7,7 @@ gem 'activesupport', '~> 4.2'
 gem 'railties', '~> 4.2'
 gem 'rails-api', '0.4.0'
 gem 'jsonite', '0.0.3'
+gem 'geokit-rails'
```
jfhamlin: This looks useful. We might be able to use IP location lookup if a client doesn't provide lat/lon.

app/controllers/resources_controller.rb
```diff
@@ -1,14 +1,22 @@
...
+    latitude = params[:latitude]
+    longitude = params[:longitude]
+
     result = if category_id
                resources.joins(:categories).where('categories.id' => category_id)
              end
 
     if result
+      if latitude and longitude
+        result = result.sort_by{ |resource| resource.address.distance_to([latitude,longitude])}
```
jfhamlin: Postgres actually comes with an extension for computing distances on the surface of the earth. We just need to install it (with a migration). http://tapoueh.org/blog/2013/08/05-earthdistance has an example of its use. We could then do this entirely in the query above with something like:

```ruby
resources.joins(:categories).joins(:address).
  where('categories.id' => category_id).
  order('point(?, ?) <@> point(address.longitude, address.latitude)',
    longitude, latitude)
```

app/controllers/resources_controller.rb
```diff
@@ -1,14 +1,22 @@
...
     category_id = params[:category_id]
+    latitude = params[:latitude]
+    longitude = params[:longitude]
+
     result = if category_id
                resources.joins(:categories).where('categories.id' => category_id)
              end
 
     if result
+      if latitude and longitude
```
jfhamlin: `&&` is usually the intended operator here. http://stackoverflow.com/questions/1426826/difference-between-and-and-in-ruby

app/presenters/resources_presenter.rb
lelandg559: will do

Gemfile
lelandg559: yea let me check into that

app/controllers/resources_controller.rb
lelandg559: ok yea that makes sense, for some reason I thought I wasn't able to include it in the db query, ill try this out

app/controllers/resources_controller.rb
lelandg559: got it

lelandg559:
Just need to add test now.

spec/requests/resources_spec.rb
```diff
@@ -31,6 +31,23 @@
...
       end
     end
+
+    context 'with a category_id and latitude/longitude' do
+      close = 10, far = 100, further = 200
+      let!(:category) { create :category, name: 'a' }
+      let!(:resources) do
+        create_list :resource, 3, categories: [category]
+      end
+      let!(:address_further) { create :address, latitude: further, longitude: further, resource_id: resources[0].id }
```
jfhamlin: I _think_ you should be able to refer to the resource directly and `FactoryGirl` will figure it out. Like `create :address, latitude: further, longitude: further, resource: resources[0]`.

spec/requests/resources_spec.rb
```diff
@@ -31,6 +31,23 @@
...
+      let!(:address_close) { create :address, latitude: close, longitude: close, resource_id: resources[1].id }
+      let!(:address_far) { create :address, latitude: far, longitude: far, resource_id: resources[2].id }
+      it 'returns only resources with that category' do
```
jfhamlin: I think we're interested in the order of the results in this test rather than that they're in the correct category

lelandg559: yea thats a copy paste mistake, will fix

jfhamlin:
LGTM

PR #88, opened 2016-06-23
Include associations in searched resource fields. Default to English.

jfhamlin:
@lelandg559, @jjfreund: Since the first bit already went out, we might as well make it work properly. This uses English as the search configuration (so that stemming works) and includes fields from several associations in the examined set.

Performance sucks, by the way, at around 100ms per query. We aren't using any indexes yet, though, and the DB instance is small, so we can worry about optimizing this later (once we're actually satisfied that search is working as we'd like).


spec/services/resources/search_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'spec_helper'
+
+describe Resources::Search do
+
```
jfhamlin: I've already got a request spec, but I'll add a few more targeted unit tests here.

jfhamlin:
@lelandg559, @jjfreund, updated with a simple unit test and some cleanup to the resources controller.

app/services/resources/search.rb
```diff
@@ -1,18 +1,34 @@
 module Resources
   class Search
-    SEARCH_COLUMNS = %i(
-      name
-      short_description
-      long_description
-      website
-    ).freeze
+    SEARCH_CONFIG = 'english'.freeze
```
jfhamlin: With this passed to `to_tsvector` and `plainto_tsquery`, Postgres will actually normalize the fields nad query with an English stemmer. My local database uses English by default, but it seems Postgres in RDS has no default configuration, so we have to specify it explicitly.

lelandg559:
LGTM

PR #89, opened 2016-06-27
Jjfreund/milestone6 domain take2

jjfreund:
Added models for users (stripped down), reviews, ratings, and keywords


db/migrate/20160626215724_create_reviews.rb
```diff
@@ -0,0 +1,8 @@
+class CreateReviews < ActiveRecord::Migration
+  def change
+    create_table :reviews do |t|
+      t.text :review
+      t.references :rating, index: true, foreign_key: true
```
jfhamlin: Should this be non-nullable?

db/migrate/20160626235137_create_join_table_category_keyword.rb
```diff
@@ -0,0 +1,8 @@
+class CreateJoinTableCategoryKeyword < ActiveRecord::Migration
+  def change
+    create_join_table :Categories, :Keywords do |t|
```
jfhamlin: `ActiveRecord` appears to normalize them, but I think [the parameters are conventionally lowercase](http://apidock.com/rails/ActiveRecord/ConnectionAdapters/SchemaStatements/create_join_table) (as the table names are).

db/migrate/20160626214555_create_ratings.rb
```diff
@@ -0,0 +1,10 @@
+class CreateRatings < ActiveRecord::Migration
+  def change
+    create_table :ratings do |t|
+      t.decimal :rating
+      t.references :user, index: true, foreign_key: true
```
jfhamlin: I think this column should be non-null.

db/migrate/20160626235137_create_join_table_category_keyword.rb
jfhamlin: Ah, `ActiveRecord` doesn't normalize them when creating the join table. These should def be lower case.

db/migrate/20160626215946_create_keywords.rb
```diff
@@ -0,0 +1,7 @@
+class CreateKeywords < ActiveRecord::Migration
+  def change
+    create_table :keywords do |t|
```
jfhamlin: If these aren't surfaced in the UI and are just to improve search quality, I'd still hold off on introducing keywords till after a basic search feature is implemented and there's a clear need.

db/migrate/20160626214555_create_ratings.rb
```diff
@@ -0,0 +1,10 @@
...
+      t.references :resource, index: true, foreign_key: true
+      t.references :service, index: true, foreign_key: true
```
jfhamlin: I'm curious if what we've got planned for the UX for ratings and reviews: what's that interaction going to look like, and are we building both in the front-end immediately?

app/models/rating.rb
```diff
@@ -0,0 +1,6 @@
+class Rating < ActiveRecord::Base
+  belongs_to :user
+  belongs_to :resource
+  belongs_to :service
+  has_one :review
```
lelandg559: should this also has_one :user ?
(the user that made the rating)

lelandg559: this was incorrect, to be clear

belongs_to will place the foreign key in the declaring model whereas has_one will place it in the other model.

we definitely want rating to have a foreign key to user , not user having a foreign key to rating.

lelandg559:
LGTM, can talk about adjusting the keywords when we get to implementing that.

Issue #77, opened 2016-06-12
Remove image path from category table


jjfreund:
Image path can be derived instead by the web app.

PR #102, opened 2016-07-19
Remove image_path from CategoryPresenter, rename presenter.

jfhamlin:
First step for #77


app/presenters/category_presenter.rb
```diff
@@ -0,0 +1,3 @@
+class CategoryPresenter < Jsonite
```
jfhamlin: Presenters are usually named with the singular resource name.

PR #103, opened 2016-07-19
[migration] Drop image_path from categories table.

jfhamlin:
This should only be merged/applied after #102.


lelandg559:
dependencies for this have been merged, nothing uses this column.

lelandg559:
verified

PR #107, opened 2016-07-21
Add rating/review creation endpoint.

jfhamlin:
@lelandg559, @jjfreund, opening this PR for discussion. I'm starting with some tweaks to the ratings table. AFAICT from the mocks in Zeplin, ratings and reviews are left on "resources"__*__. @derekfidler, is this correct?

__*__: _or OpenReferral "locations" — I'm going to get on that rename soon, before we dig ourselves even deeper on "resources"_

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,4 +2,16 @@ class ApplicationController < ActionController::API
   rescue_from ActionController::ParameterMissing do
     render nothing: true, status: :bad_request
   end
+
+  private
+
+  def require_authorization!
+    head :unauthorized unless current_user
+  end
+
+  def current_user
+    return @current_user if defined? @current_user
+    user_id = request.headers['Authorization']
+    @current_user = User.where(id: user_id).first
+  end
 end
--- a/spec/controllers/ratings_controller_spec.rb
+++ b/spec/controllers/ratings_controller_spec.rb
@@ -1,0 +1,4 @@
+require 'rails_helper'
+
+RSpec.describe RatingsController, type: :controller do
+end
--- a/spec/controllers/ratings_controller_spec.rb
+++ b/spec/controllers/ratings_controller_spec.rb
@@ -1,4 +1,0 @@
-require 'rails_helper'
-
-RSpec.describe RatingsController, type: :controller do
-end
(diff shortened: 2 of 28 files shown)


app/controllers/application_controller.rb
```diff
@@ -2,4 +2,16 @@ class ApplicationController < ActionController::API
...
+
+  private
+
+  def require_authorization!
+    head :unauthorized unless current_user
+  end
+
+  def current_user
+    return @current_user if defined? @current_user
+    user_id = request.headers['Authorization']
```
jfhamlin: This is just to get us off the ground. Once we have proper login support, we'll be providing proper auth tokens here. For now, we'll allow requests to claim to be any user by providing an existing user's ID.

spec/requests/resources/ratings/create_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+          'review' => nil
+        }
+      )
+    end
+
+    context 'with a review' do
+      let(:rating_body) do
+        {
+          rating: 3,
+          review: 'Great staff.'
```
jfhamlin: It seems to make most sense to present ratings and reviews as combined entities to clients. A single request creates a rating with an optional review.

spec/controllers/ratings_controller_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'rails_helper'
+
+RSpec.describe RatingsController, type: :controller do
```
jfhamlin: If there are no controller tests to be added, this file should be removed.

spec/requests/resources/ratings/create_spec.rb
jfhamlin: It'd be good to strip/trim review text before creating a review object. An empty string or a whitespace-only string should _not_ create a review.

jfhamlin:
@jjfreund, @lelandg559, I'm contemplating dropping the `reviews` table and just adding a nullable `review` column to `ratings`. It would make the code a bit simpler, and I don't think we'll lose any expressiveness. What do you think?

lelandg559:
ill rebase/fix this up so we can use it.

lelandg559:
after talking with joe, I will add ratings for services.

also reviewing collapsing ratings/reviews into a single table, as suggested

jjfreund:
I'll merge in when the conflict is cleared.

Issue #106, opened 2016-07-21
When importing from link-sf, set empty or poor resource (and service?) descriptions to null.


jfhamlin:
#97 calls for using service descriptions as resource descriptions. Doing this will improve the description shown in search results, but it may harm the resource page for a resource modified in this way, where a service description will appear twice. An alternative solution is to store `null` for these descriptions. If a service description would improve the UX of search results that lack resource descriptions, the client can include logic to display it in place of the missing data. This approach also has the benefit of representing missing descriptions in the database as `null`.

Issue #97, opened 2016-07-07
Correct poor descriptions in link-sf data


jjfreund:
Currently, some link-sf resource descriptions are very poor -- many are just "description." Some of these resources have better service-level descriptions. Evaluate using service level descriptions as the descriptions of these resources.

PR #108, opened 2016-07-21
Jjfreund/milestone7 miscellaneous

jjfreund:


app/presenters/ratings_presenter.rb
```diff
@@ -0,0 +1,4 @@
+class RatingsPresenter < Jsonite
+  property :rating
+  property :review, with: ReviewsPresenter
```
jfhamlin: In [my PR for ratings](https://github.com/ShelterTechSF/askdarcel-api/pull/107/files#diff-b876217aacfcd1aa62c2a771de35f8b2R4) I flattened this out, which seems a bit more straightforward, since a review is just a string. We can drop the extra presenter if we go this way.

app/presenters/ratings_presenter.rb
```diff
@@ -0,0 +1,4 @@
+class RatingsPresenter < Jsonite
```
jfhamlin: The presented type name is conventionally singular here. We strayed for some of the other presenters; I'm planning to fix that. Could we fix this one here?

lib/tasks/linksf.rake
```diff
@@ -42,10 +42,19 @@ namespace :linksf do
...
+
+      puts 'resource description is :' + resource.long_description
+
       record[:services].each do |service_json|
         category_name = service_json[:category]
         category_name = 'shelter' if category_name == 'housing'
         cat = Category.where('lower(name) = ?', category_name).first
+
+        if resource.long_description.blank? || resource.long_description.length < 15
+          puts 'replacing bad description' + resource.long_description + ' with ' + service_json[:description]
```
jfhamlin: NBD since this is just informational, but there's a missing space after "bad description".

Also, it looks like we're not doing any replacing here, just printing this out. Might be worth bringing up #106 as an implementation of #97 if the two approaches are of comparable implementation cost.

app/presenters/ratings_presenter.rb
```diff
@@ -0,0 +1,4 @@
...
+  property :rating
```
jfhamlin: Ah, I also noticed that this ends up getting returned as a string. I'm not sure why. I had to cast to a float with `to_f` when I was working on this. I assume that's better for the web client. Which reminds me: it'd be good to update the resource request test(s) to include some ratings.

jfhamlin:
LGTM

Issue #74, opened 2016-06-07
Populate factory girl mock categories with image icons


jjfreund:

jfhamlin:
Instead of storing the icon filenames in the database, can the clients map category names to icon names by convention? It looks like `'ic-' + category.name.toLower().replace(' ', '-') + '@3x.png'` would work. And if it doesn't we can rename them so it does or use some other mapping. The files are in the web app anyway, so I don't think we gain much by returning icon names from the API.

PR #109, opened 2016-08-15
Search for resources doesn't order results by distance #74

lelandg559:
Need some help with this, I can add this so that it does the sorting but I’m having trouble with that syntax of keeping the distinctness in the query,

I believe this would be done with a subquery but I’m not sure of the syntax.

### DIFF
--- a/app/services/resources/search.rb
+++ b/app/services/resources/search.rb
@@ -22,14 +22,18 @@ module Resources
       end
     end.flatten.join('||').freeze
 
-    def self.perform(query, scope: Resource)
+    def self.perform(query, sort, scope: Resource)
+      sort = 'resources.name' unless sort
+
       scope
         .joins('LEFT JOIN services ON services.resource_id = resources.id')
         .joins('LEFT JOIN notes ON notes.resource_id = resources.id')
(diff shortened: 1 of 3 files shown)


app/services/resources/search.rb
```diff
@@ -22,14 +22,15 @@ class Search
...
+    def self.perform(query, scope: Resource, sort: sort_order)
       scope
         .joins('LEFT JOIN services ON services.resource_id = resources.id')
         .joins('LEFT JOIN notes ON notes.resource_id = resources.id')
         .joins('LEFT JOIN categories_resources ON categories_resources.resource_id = resources.id')
         .joins('LEFT JOIN categories ON categories.id = categories_resources.category_id')
+        .joins('LEFT JOIN addresses ON addresses.resource_id = resources.id')
         .where("#{CLAUSE} @@ plainto_tsquery('#{SEARCH_CONFIG}', ?)", query)
-        .distinct
+        .order(sort)
```
lelandg559: Need to keep the distinctness here but having both on the same query results in 
ActiveRecord::StatementInvalid (PG::InvalidColumnReference: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list

Need to do this as a subquery I believe?

PR #111, opened 2016-09-06
Remove image_path from CategoryPresenter, rename presenter. #102

lelandg559:


lelandg559:
front end isnt set up to handle this yet, will look into that.

PR #115, opened 2016-12-06
Added devise and admin role

lelandg559:


lelandg559:
Already in master via another branch.

PR #118, opened 2016-12-15
docker-compose development flow

richardxia:
I prefer a Docker-based development workflow, since it solves a lot of problems with making sure the right versions of different requirements are installed. I've added a `docker-compose.yml` file that orchestrates a two-container setup, with one database container and one web (Rails) container.

I couldn't get the existing Dockerfile to work. Here's the error I was getting when trying to build it:

```
Building web
Step 1 : FROM ad2games/docker-rails:2.1.1
# Executing 6 build triggers...
Step 1 : COPY Gemfile Gemfile.lock /home/app/webapp/
 ---> Using cache
Step 1 : RUN chown app:app Gemfile Gemfile.lock &&   chpst -u app bundle install --deployment --jobs 4 --without development test &&   find vendor/bundle -name *.gem -delete
 ---> Using cache
Step 1 : COPY . /home/app/webapp/
Step 1 : RUN mkdir -p db public/assets log tmp vendor
 ---> Running in e30d22f6da71
Step 1 : RUN chown -R app:app app db public log tmp vendor
 ---> Running in 5259b9315102
Step 1 : RUN bash -ec   "if chpst -u app bundle show sprockets; then     touch .env; source .env;     chpst -u app bundle exec rake assets:precompile       SECRET_KEY_BASE=noop DATABASE_URL=postgres://noop ;   fi"
 ---> Running in 29b1f251a93a
/home/app/webapp/vendor/bundle/ruby/2.3.0/gems/sprockets-3.7.0
rake aborted!
PG::ConnectionBad: could not translate host name "noop" to address: Name or service not known
```

Since I didn't want to mess with the existing Dockerfile, in case if that's how the production site is run, I used the `ruby:2.3.0-onbuild` image, which functions similarly enough. Let me know if there's a simple way to fix my problem, and I'll gladly get rid of the second Dockerfile I added.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -24,6 +24,7 @@ default: &default
 development:
   <<: *default
   database: askdarcel_development
+  url: <%= ENV['DATABASE_URL'] %>
 
   # The specified database role being used to connect to postgres.
   # To create additional roles in postgres see `$ createuser --help`.
@@ -58,6 +59,7 @@ development:
 test:
   <<: *default
   database: askdarcel_test
+  url: <%= ENV['TEST_DATABASE_URL'] %>
 
 # As with config/secrets.yml, you never want to store sensitive information,
 # like your database password, in your source code. If your source code is
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,19 @@
+version: "2"
+services:
+  db:
+    image: "postgres:9.5"
+
+  api:
+    build:
+      context: .
+      dockerfile: Dockerfile.dev
+    command: rails server --port=3000 --binding=0.0.0.0
+    depends_on:
+      - db
+    environment:
+      DATABASE_URL: postgres://postgres@db/askdarcel_development
+      TEST_DATABASE_URL: postgres://postgres@db/askdarcel_test
+    ports:
+      - "3000:3000"
+    volumes:
+      - .:/usr/src/app
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,1 @@
+FROM ruby:2.3.0-onbuild
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # askdarcel-api [![Build Status](https://travis-ci.org/ShelterTechSF/askdarcel-api.svg?branch=master)](https://travis-ci.org/ShelterTechSF/askdarcel-api)
 
-## Set-up Instructions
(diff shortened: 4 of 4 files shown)


config/database.yml
```diff
@@ -24,6 +24,7 @@ default: &default
 development:
   <<: *default
   database: askdarcel_development
+  url: <%= ENV['DATABASE_URL'] %>
```
richardxia: See http://edgeguides.rubyonrails.org/configuring.html#connection-preference on what Rails does if you provide both a URL and other attributes for the database, but the short answer is that it merges both sets of information, favoring the URL if there are conflicts. It is also totally cool with an empty string for the URL, so this shouldn't affect people who aren't using the environment variable.

lelandg559:
Cool, I like this,

So the problem with the existing Dockerfile has been patched, (precompiling assets is causing a gem to want a db connection which the ad2games scripts don't allow).

But given that the ad2games image is designed for running in production, it would be good to have this dev file on here.  

Tested it out and it works fine for me.

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+version: "2"
+services:
+  db:
+    image: "postgres:9.5"
```
lelandg559: There would need to be a port mapping here to allow a db client outside of the container to connect to it, right?

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+
+  web:
```
lelandg559: I might rename this as "api" , web typically refers to all the components from the askdarcel-web project.

config/database.yml
lelandg559: So if I understand correctly, this line effectively does nothing?  At best, it's just there to make it clear that rails is looking at that env variable for db url?

richardxia: Yes, it just makes it explicit. It also makes it so that the development environment parallels both the test and production environments.

docker-compose.yml
richardxia: Right. To be precise, there are two separate containers here, `db` and `web`, and any container listed here will by default be able to talk to any other container listed here. In order to talk to outside containers or things outside of Docker, then you need to expose the port.

docker-compose.yml
richardxia: Got it, I renamed it and updated the README as well.

lgarofalo:
LGTM

PR #119, opened 2016-12-19
Removed dependency on external database for ad2games rails docker image precompile

lgarofalo:


lelandg559:
I will follow this up with a commit to make the RDS instances not publicly accessible.

Issue #110, opened 2016-08-25
Clean up initial import of phone number data


achadha:
Some of the phone numbers that we're retrieving have extra characters or the like. For example, Resource 180 has an extra '.' character at the end of its phone number.

richardxia:
@jjfreund Did you mean to close this? https://github.com/ShelterTechSF/askdarcel-api/pull/124 hasn't been merged yet.

PR #124, opened 2017-01-05
Clean up phone number data

richardxia:
Refs #110.

### Overview
I found a library, [phonelib](https://github.com/daddyz/phonelib), which seems like a pretty good Ruby library for doing phone number validation. It wraps Google's [libphonenumber](https://github.com/googlei18n/libphonenumber) library/database, which is good because it receives regular updates as countries update their phone number formats and numbering allocations. I'm using this library to both normalize our existing numbers and to standardize the format in which the API returns phone numbers (using the "national format", which in the US is (XXX) XXX-XXXX).

### Updated linksf.json with clean numbers
I ran a one-off Ruby script (not included in PR) to parse the phone numbers in linksf.json and reformat it into [E.164 format](https://en.wikipedia.org/wiki/E.164), which is I think the closest thing there is to a canonical, international phone number format. The US phone number `(555) 123-4567` would map to `+15551234567` in E.164 format.

There were only two categories of numbers in the linksf database that couldn't be parsed by `phonelib` as they are currently formatted:

1. Some numbers have an extension that doesn't match the expected E.164 format for extensions (e.g. `(555) 555-5555;ext=555`). I wrote a small regex to convert these into a parseable format.
2. One number is missing the area code. I had the script assume that any 7-digit (US) number should be paired with the 415 area code, which seems to be correct for this particular number, but obviously isn't correct in general.

### Updated PhonesPresenter to use `phonelib` for formatting
Rather than using the `extension` and `country_code` fields on the Phone model, the PhonesPresenter now uses `phonelib` to parse the now E.164 formatted number to extract both of those fields.

Aside: "Country code" isn't really the proper term for what the API returns, since it is used in the E.164 specification to refer to the 1-3 digit prefix used when dialing an international number. For the US, Canada, and many other North American countries, this is the number 1. The `phonelib` library calls `US` the "country" , while the Google library calls it the "region code". Any opposition to renaming our field `region_code`, or even `iso3166_alpha2` (https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)?

### Removed `extension` and `country_code` columns

Re: description above, I added a migration that drops these two fields, since they've been folded into the `number` field.

### Updated Phone factory

The Phone factory no longer generates `extension` and `country_code` fields, and folds them into the `number` field. I had a little bit of trouble with getting the Faker library to always generate valid numbers, so I had it retry until it gets a valid one. When doing a test on 10,000 fake numbers, about 9500-9600 are valid, so it should very rarely retry.


richardxia:
Hey @jjfreund, I tried to look for a way to redump the linksf.json file, but couldn't find anything. The only thing in linksf.rake is an import task, but not an export task. Can you point me to how I can dump it out again?

richardxia:
Hey @jjfreund, I just pushed a few commits that update linksf.json, the Phone factory, and add a migration that drops the `extension` and `country_code` columns. I also git rebased away the Rake task that I had originally written, since linksf.json just has the correctly formatted phone numbers now. The PR description has been updated to reflect the current set of changes.

I think this is ready to merge, unless if there's anything else you think I should change.

richardxia:
Note to self: Revert the linksf.json file back to what it used to be, and then update the linksf import script to massage the phone numbers before saving them to the DB.

richardxia:
Okay, I reverted the changes I made to the linksf JSON file and added the number parsing code to the linksf import Raketask, so we should be able to update the linksf data and still get the cleaned numbers.

lgarofalo:
Nice, LGTM add validation, opened #138 to address validation issues.

PR #135, opened 2017-02-10
Fix Postman tests

### LINKED ISSUE #126: Fix false negatives in postman
Some tests are failing due an arbitrary performance metric; others are failing due to an arbitrary status code check.

richardxia:
Fixes #126. This fixes the Postman tests so that they *almost* all pass.

### Test fixes
- Changed expected status code for successful POSTs to 201 CREATED, since that's what they actually return and it's more correct than 200 OK
- Changed the second Admin test to operate on a different ChangeRequest, since you can't reject a ChangeRequest that's already been approved and the database isn't reset between tests
- Added an actual test assertion to the Admin signout test
- Fixed a real performance bug with the ChangeRequest index view, where nested resources weren't getting prefetched, causing O(N) database queries

### Cosmetic changes
- The Postman tests for some reason were sending a request body with a JSON-formatted rating for all GET requests, which was kind of confusing. None of the tests have to do with ratings, and normally GET requests don't even look at the request body because parameters are supposed to come from URL parameters. I removed all the request bodies from GET requests.
- There's a second, older copy of the ChangeRequestsController named `changerequests_controller.rb`, which isn't actually used. This confused me for a long time because this copy actually correctly prefetches the nested resources, and I had a hard time why I wasn't getting O(N) queries when I ran the code manually in the console. I copied that part into the real controller and deleted the rest.

### Remaining failing test
The test that checks the ChangeRequests index still fails because it doesn't run in the 1000ms time limit. This is because on a fresh linksf import, there are ~450 pending change requests. I'm guessing for testing, we'll want a specially set up database and not the linksf imported DB.

I also get a test failure on the first two tests in the non-admin collection on the first run, but not subsequent runs. They don't finish within the time limit, but I think it's just due to a lot of one-time-cached Ruby objects getting loaded.


lgarofalo:
Ran locally, this LGTM

PR #142, opened 2017-02-23
Update linksf import task.

richardxia:
It was relatively straightforward to update the import task to match the new schema. Every one of the old fields that we used exists in the new schema, so all it took was to update the paths for accessing the fields.

I also uncovered a few bugs in the old import task:

- All zip codes were fake because they were calling a fake zip code generator. Zip codes are currently required, so I just assigned them all to the empty string. The frontend is apparently already equipped to handle this, so it doesn't even add the extra comma after the state if there is no zip code.
- We were grabbing notes from the resource rather than the service, since the old schema had a field named `notes` on both resources and services. We used to use the resource's notes field as the long description, so I assumed this was in error. I'm properly grabbing the notes from the services now, so we should try to display it in the frontend. Also, note that they renamed the services' notes field to `application_process`.
- This last one I didn't fix, but I noticed that we're rounding all the schedule hours down to the hour, meaning that we lose the resolution of minutes. This means that we're misrepresenting the opening hours if they aren't exactly on the hour. E.g. if a service really opens at 8:30, we're rounding down and saying it opens at 8.


richardxia:
Related to my last bullet point, is there a reason we decided to use integers to represent the opens_at/closes_at field instead of the SQL time type?

richardxia:
Oh, and one last comment: I think the only new fields that wasn't previously available is the `eligibility` field, which appears to track age and gender requirements for the service. Should we be capturing these somehow? This is their schema: https://github.com/zendesk/linksf/blob/d63ab2d52e1863092c190c1a1e9202c5a9e1b123/lib/eligibilities.js

lgarofalo:
Integers vs SQL time data type is likely just an oversight.  Don't see a reason to not use SQL time type if the change updates everything using it appropriately.

lgarofalo:
LGTM and verified.

PR #147, opened 2017-03-16
145 apicaching

lgarofalo:


richardxia:
I figured out what I was doing "wrong" after I typed this up, but it's still something that could affect real users so I'm going to post it as I originally wrote it:

---

@lgarofalo, I'm testing this branch locally, but I'm finding that the caching isn't helping because the API calls end up being slightly different due to the longitude and latitude changing ever so slightly. This is on a laptop, so I'm not sure why the coordinates are changing so much, but it's what I'm seeing. Here's some example API request params that I'm seeing:

```
/resources?category_id=1&lat=37.751797599999996&long=-122.39948910000001
/resources?category_id=1&lat=37.7517649&long=-122.39953279999999
/resources?category_id=1&lat=37.751775599999995&long=-122.3995202
/resources?category_id=1&lat=37.7517649&long=-122.3994763
```

Maybe we should do a bit of rounding in the frontend code before making API requests? A lot of these look like floating point precision issues. The DB seems to only save up to 7 decimal places.

---

I figured out the reason my coordinates were changing were because I kept on hitting the "Ask Darcel" logo to go back to the home page, which causes the browser to actually do a real page load instead of just updating the React state. This causes the app to make a new query for the location, which apparently can slightly change even when on a non-mobile device. I'll file a ticket on the web repo to do the routing in React instead of causing a real page load when you hit the logo.

richardxia:
I moved the `before_action :cachable` out of ApplicationController and into the individual Categories and Resources controllers, per the discussion we had IRL about it being confusing if an admin approved a change request but it still showed up on the change requests page. This also takes care of the problem of caching non-read-only routes, since the Categories and Resources controllers only have read-only controllers.

lgarofalo:
Read your comments.

Will file an issue about the GPS rounding, 
Noticed the react reload off logo click as well.

I'll go through and update the cache-able calls with the right tag and verify.

20 minutes is still arbitrary for I think that is sufficient for user experience for now.

PR #148, opened 2017-03-16
Add services

lgarofalo:


postman/AskDarcel%20API.postman_collection.json
```diff
@@ -444,6 +465,44 @@
...
+					},
+					{
+						"key": "Content-Type",
+						"value": "application/json",
+						"description": ""
+					}
+				],
+				"body": {
+					"mode": "raw",
+					"raw": "{\n    \"rating\": {\n    \t\"rating\": {\n    \t\"user\": {\n    \t\t\"id\": 1\n    \t},\n    \t\"rating\": 3\n    }\n    }\n}"
```
richardxia: @lgarofalo @achadha, what's the correct POST body for creating a new service request? The one in the Postman JSON doesn't look right:

```json
{
    "rating": {
        "rating": {
            "user": {
                "id": 1
            },
            "rating": 3
        }
    }
}
```

richardxia:
@lgarofalo, I think this branch is ready for review now. I worked with @achadha to make sure the create endpoint works for him in the format he expects.

Here are the changes I added on to what you had done:

- Implemented the create method, which now takes a list of services and validates and creates them all.
- Set the default status to pending
- Updated the migration you added to initialize all existing services' statuses to approved. For a local dev env, if you already ran your migration, you'll need to `rake db:rollback && rake db:migrate` to pick up the new change, but this should do the right thing when we push to prod.
- Updated the resource controller to only include approved services (#149)
- Updated the Postman test for creating services, and added a small RSpec test for checking that only approved services are returned

PR #153, opened 2017-03-25
Removed caching from API for now

lgarofalo:


lgarofalo:
Doing this for short term issue.

PR #164, opened 2017-04-17
Adjust cacheable to 5 minutes and add to Categories and Resources

lgarofalo:


lgarofalo:
These changes were previously approved.

PR #168, opened 2017-05-21
Algolia Search Backend

BenjaminCorey:
## What is this? Why?

Search backed by Algolia, allowing fuzzy matching and configuration of search ranking through algolia's backend.

- [x] Indexing for Resources 
- [x] Configure `/resources/search` to use Algolia
- [x] Rake task for for initial production indexing
- [x] Add `touch: true` to nested models so that they trigger an update of the Resources index.
- [ ] Setup production Algolia account
- [ ] Strategy for managing secrets? Currently I've stashed them in `.env` locally, not sure how we're doing this on production.
- [ ] Testing (mock algolia?)

@jjfreund please take a look.

Feature complete, needs tests, review and QA.

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -16,7 +16,7 @@ class ResourcesController < ApplicationController
 
   def search
     result = Resources::Search.perform(params.require(:query), sort_order, scope: resources)
-    render json: ResourcesPresenter.present(result)
+    render json: {resources: ResourcesPresenter.present(result)}
   end
 
   private
--- a/app/models/resource.rb
+++ b/app/models/resource.rb
@@ -1,4 +1,8 @@
 class Resource < ActiveRecord::Base
+  include AlgoliaSearch
+
+  delegate :latitude, :longitude, to: :address, prefix: true, allow_nil: true
+
   has_and_belongs_to_many :categories
   has_and_belongs_to_many :keywords
   has_one :address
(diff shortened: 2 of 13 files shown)


app/controllers/resources_controller.rb
```diff
@@ -16,7 +16,7 @@ def show
 
   def search
     result = Resources::Search.perform(params.require(:query), sort_order, scope: resources)
-    render json: ResourcesPresenter.present(result)
+    render json: {resources: ResourcesPresenter.present(result)}
```
BenjaminCorey: I'm not sure why I had to nest this explicitly, shouldn't the presenter take care of this? Maybe because the return from the search is an Array of resources and not an ActiveRecord Association?

docker-compose.yml
```diff
@@ -13,6 +13,8 @@ services:
     environment:
       DATABASE_URL: postgres://postgres@db/askdarcel_development
       TEST_DATABASE_URL: postgres://postgres@db/askdarcel_test
+      ALGOLIA_APPLICATION_ID:
+      ALGOLIA_API_KEY:
```
BenjaminCorey: These keys are required now, so you'll have to sign up for Algolia and create a new 'application', then copy/paste the values into your local `.env` file.

jjfreund: Is there a way we could have people use a shared application id? Ideally, developers wouldn't have to manually configure anything in their environment files.

app/models/resource.rb
```diff
@@ -8,4 +12,39 @@ class Resource < ActiveRecord::Base
   has_many :services
   has_many :ratings
   has_many :change_requests
+
+  algoliasearch per_environment: true do
```
jjfreund: Does this have to be in the model? In general I like to keep model objects as spare as possible, as it keeps the application more orderly and often makes things easier to test, but in this case I see why it might be easiest. Could you perhaps have a separate search object that wraps the Resource, maybe?

docker-compose.yml
richardxia: Or alternatively, could we fallback to a naive search if the credentials aren't available? This would also be useful for developing without an available internet connection and for the testing environment, which probably shouldn't be making real requests to the internet.

jjfreund: Also a good idea.

app/models/resource.rb
```diff
@@ -8,4 +12,39 @@ class Resource < ActiveRecord::Base
...
+          postal_code: address.postal_code,
+          country: address.country,
+        }
+      else
+        {}
+      end
+    end
+
+    add_attribute :notes do
+     notes.map {|n| n.note }
```
richardxia: Ruby trick: You can use `&:note` as shorthand for `{|n| n.note}`, so this could just be written as `notes.map(&:note)`. Same with the other attributes below.

lgarofalo:
From Travis, ArgumentError: No APPLICATION_ID provided, please set :application_id

Can be set here https://travis-ci.org/ShelterTechSF/askdarcel-api/settings

richardxia:
Superseded by #179.

PR #177, opened 2017-07-03
Jjfreund/make changerequests editable

jjfreund:
Allows for the modification of change requests.

### DIFF
--- a/app/controllers/change_requests_controller.rb
+++ b/app/controllers/change_requests_controller.rb
@@ -43,6 +43,21 @@ class ChangeRequestsController < ApplicationController
     else
       change_request = ChangeRequest.find params[:change_request_id]
       if change_request.pending?
+
+        puts 'eh'
+        puts params[:change_request]
+        puts 'hi'
+
+        puts change_request.id
+
+        puts 'wut'
+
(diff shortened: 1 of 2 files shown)


app/controllers/change_requests_controller.rb
```diff
@@ -54,6 +61,13 @@ def approve
     end
   end
 
+  def replace_field_changes(change_request)
+
+
+
+    return change_request
+  end
+
```
richardxia: Is this method used anymore? It also doesn't seem to be doing anything now.

app/controllers/change_requests_controller.rb
```diff
@@ -43,6 +43,13 @@ def approve
     else
       change_request = ChangeRequest.find params[:change_request_id]
       if change_request.pending?
+
+        FieldChange.delete_all(["change_request_id = ?", change_request.id])
+
+        change_request.field_changes = field_changes_approve change_request.id
+
+        change_request.save!
```
richardxia: For safety, perhaps surround this set of changes with a transaction so that in case if an error occurs, we don't end up deleting the old field changes without creating the new ones?

PR #179, opened 2017-07-05
Algolia Search Backend (new PR)

richardxia:
cc @BenjaminCorey, who I apparently can't assign as a reviewer yet

I had to create a new pull request because https://github.com/ShelterTechSF/askdarcel-api/pull/168 was created on Ben's fork of this repo, which I don't have push permissions to.

All I did on top of https://github.com/ShelterTechSF/askdarcel-api/pull/168 was add the old database-based search back in and conditionally chose between the Algolia search and the database search by checking whether the Algolia credentials were present. I also cleaned up a bit of the linting to appease rubocop.

I don't know the answer to "Strategy for managing secrets? Currently I've stashed them in .env locally, not sure how we're doing this on production," can @lgarofalo comment on how secrets are injected in production? I don't know if they're normally baked into the container image or if ECS has some way of injecting them at container startup time.


app/models/resource.rb
```diff
@@ -8,4 +12,41 @@ class Resource < ActiveRecord::Base
   has_many :services
   has_many :ratings
   has_many :change_requests
+
+  if Rails.configuration.x.algolia.enabled
```
BenjaminCorey: @jjfreund had mentioned something about pulling this out into a module. Although, that might be even more confusing b/c algolia registers AR lifecycle events.

richardxia: Right, I had the same train of thought. I figured it'd probably be better give in to the way `algolia` expects you to integrate it, since it's less error-prone.

jjfreund: Fair enough.

config/application.rb
```diff
@@ -9,5 +9,13 @@
 module AskdarcelApi
   class Application < Rails::Application
     config.api_only = true
+
+    # Algolia
+    config.x.algolia.application_id = ENV['ALGOLIA_APPLICATION_ID']
+    config.x.algolia.api_key = ENV['ALGOLIA_API_KEY']
```
lgarofalo: https://github.com/ShelterTechSF/cloudformation/pull/23

The template will have placeholders in repo but secrets will be set in AWS

lgarofalo:
Do we need to configure env variable values to merge this safely?

lgarofalo:
> Conflicting files
> Gemfile
> Gemfile.lock
> app/models/resource.rb
> app/models/service.rb
> app/services/resources/search.rb
> config/application.rb

😢 , I'll take a look at resolving that and getting this merged unless anyone else is already doing that.

richardxia:
I tested this against an Algolia trial account and can confirm that it's still working. We'll still need the "real" account API keys to use this in prod (and maybe staging as well). Alternatively, we can call my account the "real" one and I can try to upgrade it to the free [open source](https://www.algolia.com/for-open-source) plan.

> Do we need to configure env variable values to merge this safely?

If we don't want the environment variables to block merging this, I can make a quick change to allow production to fallback to the old, non-Algolia-backed search by removing this block: https://github.com/ShelterTechSF/askdarcel-api/blob/395efa218921e6fa8809b66588d4ef0cbe6ff218/config/application.rb#L17-L19

lgarofalo:
@richardxia Let's add the fallback and I can update the environments with the secret once we release this.

Worst case we remove the secret and it falls back.

richardxia:
Sounds good, I just removed the code that requires production to have the Algolia credentials. Should be good to merge now.

richardxia:
> Alternatively, we can call my account the "real" one and I can try to upgrade it to the free open source plan.

I contacted @BenjaminCorey and he let me know that he had not upgraded his account, so I'll go send Algolia a request to upgrade my account and we can make that the official prod account.

PR #180, opened 2017-07-06
Allow adding change request for new phone number.

richardxia:
Here I'm proposing that we leverage ChangeRequests to support creating new model instances, as opposed to only supporting changes to already existing models. It looks like the `object_id` field on ChangeRequests is already nullable, so all I had to do was update the controller logic to create a new object if `object_id` is nil.

I had to add two new, backwards-compatible parameters to the API:
- `type`, which should be the (plural) name of the model being changed
- `parent_resource_id`, which should be the ID of the Resource that the model belongs to

`type` was necessary because we can no longer rely on the presence of `phone_id` to let us know the type of model being created. `parent_resource_id` was necessary because 1) we can't look up the Resource via the Phone, since one does not already exist in this case, and 2) `resource_id` was already used as a parameter name to imply that the proposed change request should be made on the resource itself.

These API changes are backwards compatible, since changes requests for model instances that already exist will use the existing codepaths.

I only made the API change for phone numbers. It's going to require a bit more work to have this scheme apply to other types of models, since we may run into issues with database column constraints. Phone's `service_type` field is not nullable, which means that we need to initialize it to the empty string prior to applying the field changes.

I have another branch that I'm working on to actually add in this new API call in the frontend, but from an API perspective, these changes are sufficient for me to make an edit and approve it via the admin page.

### DIFF
--- a/app/controllers/change_requests_controller.rb
+++ b/app/controllers/change_requests_controller.rb
@@ -37,7 +37,7 @@ class ChangeRequestsController < ApplicationController
     if !admin_signed_in?
       render status: :unauthorized
     else
-      render json: ChangeRequestsPresenter.present(changerequest.pending)
+      render json: ChangeRequestsWithResourcePresenter.present(changerequest.pending)
     end
   end
 
--- a/app/controllers/change_requests_controller.rb
+++ b/app/controllers/change_requests_controller.rb
@@ -6,8 +6,9 @@ class ChangeRequestsController < ApplicationController
       @change_request = ServiceChangeRequest.create(object_id: params[:service_id], resource_id: Service.find(params[:service_id]).resource_id)
     elsif params[:address_id]
       @change_request = AddressChangeRequest.create(object_id: params[:address_id], resource_id: Address.find(params[:address_id]).resource_id)
-    elsif params[:phone_id]
-      @change_request = PhoneChangeRequest.create(object_id: params[:phone_id], resource_id: Phone.find(params[:phone_id]).resource_id)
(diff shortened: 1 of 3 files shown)


jjfreund:
The new workflow is far superior for simple objects, great work there. Unsure if we need backwards compatibility but I suppose it can't hurt. Could you include a postman test?

richardxia:
Postman test added: https://github.com/ShelterTechSF/askdarcel-api/pull/180/commits/6e1939cd0a29aa20d060e5cb80b9380b6423a49b

richardxia:
So apparently adding my test uncovered an O(N) queries problem with the change requests controller, unrelated to the actual code I changed. AFAICT, the only reason my test fails is because the O(N) queries are on the Resource -> Services relationship, and it's because my test runs *after* a different test that adds a second Service to the Resource with id 1.

Question: Can we split ChangeRequestsPresenter into two different presenters? It currently recursively presents the associated resource, which I believe was only added for the admin view, but it's completely ignored by the edit view. In fact the edit view doesn't do anything with the response other than checking the status code.

Followup question: I feel like this is some code smell w.r.t. the modeling hierarchy. Should ChangeRequests really "contain" their associated Resource? Or should it be the other way around, with Resources containing all of their ChangeRequests?

richardxia:
Also, the reason I didn't catch this before pushing the Postman test to Travis was because the `bullet` gem is only enabled in the `test` environment, and I guess Travis automatically sets the Rails environment to `test`. Should we just enable Bullet in the development environment as well?

PR #181, opened 2017-07-12
Jjfreund/adding resources

jjfreund:


app/controllers/resources_controller.rb
```diff
@@ -19,8 +19,61 @@ def search
     render json: ResourcesPresenter.present(result)
   end
 
+  def create
+    resources_params = clean_resources_params
+    resources = resources_params.map { |r| Resource.new(r) }
```
richardxia: Do we need this route to accept a list of resources? I think this is a little bit unexpected because the standard Rails CRUD routes only allow you to create one object per request.

spec/factories/resources.rb
```diff
@@ -1,6 +1,7 @@
 FactoryGirl.define do
   factory :resource do
     name { Faker::Company.name }
+    status 1
```
richardxia: Are you able to use the symbol name? That'd be a lot clearer than the number 1.

app/controllers/resources_controller.rb
```diff
@@ -4,7 +4,7 @@ class ResourcesController < ApplicationController
   def index
     category_id = params.require :category_id
     relation = resources.joins(:categories).joins(:address)
-                        .where('categories.id' => category_id)
+                        .where('categories.id' => category_id).where('status' => :approved)
```
lgarofalo: This doesn't map correctly to value in database.

WHERE "categories"."id" = $1 AND "resources"."status" = $2        
           [["id", 1], ["status", :approved]]
I, [2017-09-05T02:41:38.886836 #48]  INFO -- : Completed 500 Internal Server Error in 8ms (ActiveRecord: 1.7ms)
F, [2017-09-05T02:41:38.887332 #48] FATAL -- :   
F, [2017-09-05T02:41:38.887367 #48] FATAL -- : ActiveRecord::StatementInvalid (PG::InvalidTextRepresentation: ERROR:  invalid input syntax for integer: "approved"

PR #191, opened 2017-09-06
fix(api): update get resources by category to use correct status enum…

lgarofalo:
… value

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -5,7 +5,7 @@ class ResourcesController < ApplicationController
     category_id = params.require :category_id
     relation = resources.joins(:categories).joins(:address)
                         .where('categories.id' => category_id).where(status: Resource.statuses[:approved])
-    .order(sort_order)
+                        .order(sort_order)
     render json: ResourcesPresenter.present(relation)
   end
 
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -6,7 +6,7 @@ class ResourcesController < ApplicationController
     relation = resources.joins(:categories).joins(:address)
                         .where('categories.id' => category_id).where(status: Resource.statuses[:approved])
     .order(sort_order)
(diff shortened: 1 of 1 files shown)


app/controllers/resources_controller.rb
```diff
@@ -4,9 +4,9 @@ class ResourcesController < ApplicationController
   def index
     category_id = params.require :category_id
     relation = resources.joins(:categories).joins(:address)
-                        .where('categories.id' => category_id).where('status' => :approved)
-                        .order(sort_order)
-    render json: ResourcesPresenter.present(relation)
+                        .where('categories.id' => category_id).where(status: Resource.statuses[:approved])
+    .order(sort_order)
+    render json: ResourcesPresenter.present(relation.approved)
```
lgarofalo: that was unintended, fixing.

lgarofalo:
@jjfreund: "ship it"

app/controllers/resources_controller.rb
```diff
@@ -4,7 +4,7 @@ class ResourcesController < ApplicationController
   def index
     category_id = params.require :category_id
     relation = resources.joins(:categories).joins(:address)
-                        .where('categories.id' => category_id).where('status' => :approved)
+                        .where('categories.id' => category_id).where(status: Resource.statuses[:approved])
```
richardxia: I don't understand why this wasn't working before. The [ActiveRecord docs](http://api.rubyonrails.org/classes/ActiveRecord/Enum.html) explicitly say that you can do this:

```rb
Conversation.where(status: [:active, :archived])
Conversation.where.not(status: :active)
```

lgarofalo: Those were working for me as well, to be safe I went with the most explicit version I saw that worked.

Interestingly though, 'where('status' => :approved)' works in the dev environment but not on staging.

¯\_(ツ)_/¯

My only suspicion currently is the difference in pre-compiling ruby before it goes to staging, otherwise versions are the same.

PR #195, opened 2017-09-09
Add new categories and mark some as top_level categories.

richardxia:
## Deployment notes:

I have a front-end PR that needs to be merged in before this, or else you'll flood the home page with 163 categories: https://github.com/ShelterTechSF/askdarcel-web/pull/302

Also, we need to run the one-time rake task `rake onetime:add_categories` after the deploy in order to actually populate the new categories. It's idempotent, so it's safe to run it multiple times.

## Changes
- Add boolean top_level field to Category
- Mark existing Categories as top level
- Add rake task to add new Categories and set top-level field appropriately
- Add parameter to CategoriesController to filter by top_level:
    - If unset or set to empty string, do not do any filter
    - If set to false, False, f, 0, etc, return only non-top-level categories
    - Otherwise return only top-level categories
- Update linksf and db:populate tasks to set initial category fixtures as
  top-level

## Screenshots
screen shot 2017-09-10 at 6 15 33 pm
screen shot 2017-09-09 at 4 37 11 pm

## Open questions

@derekfidler, regarding the new top-level categories you sent me, I noticed there is almost no overlap between the old categories and the new ones. The only one that is still in the new list is "Food". Shelter, Medical, Hygiene, and Technology are now not top-level, which means that we need to do a lot of cleanup of the data to actually be able to see results when clicking the category buttons on the front page.


richardxia:
Copying the notes from Slack here regarding the migration plan:

Rename Shelter to Housing
Rename Medical to Health
Create a new category named Homelessness Essentials and take every Resource and Service originally tagged with Hygiene and also tag it with Homelessness Essentials. These Resources and Services will be tagged with both categories
Technology should also be treated the same, as in tag all of the existing Resources and Services with Homelessness Essentials as well

I'm going to make these changes to the one-time rake task now.

richardxia:
I think this is ready for final review. I got the last set of changes in which apply the new migration strategy for the existing categories. Here's the new icon for the front page:

screen shot 2017-09-10 at 6 15 33 pm

One minor issue I discovered is that we [hardcoded category names](https://github.com/ShelterTechSF/askdarcel-web/blob/733364941875c97a1349691a16b453587be33a0f/app/components/Search/ResourcesTable.js#L13-L19) in one place the frontend code, so now the category title is wrong or missing on the search results page:

screen shot 2017-09-10 at 6 49 46 pm

For the categories that got renamed, it'll display their old, hardcoded name. I didn't want to continue hardcoding more categories into the frontend, especially because we won't know what IDs they'll have until we actually run the migrations in prod, so I think we should just deploy without fixing this problem and create a separate ticket for cleaning up that code later.

PR #202, opened 2017-09-14
Fix add categories change request

richardxia:
This fixes the other issue with ShelterTechSF/askdarcel-web#307, where changes to a service's categories weren't getting persisted. The issue is that we don't actually have an API for being able to set many-to-many relationships via a change request. The ChangeRequest and FieldChange models assume that changes are always to single values, not arrays of values.

In order to make this work, I've added some hacky code paths to check if the name of the field you're changing is `categories`, and if so, serialize the array into a string via its JSON encoding. Then in the `#persist_changes` method, I deserialize the array so that it can be passed into `Model#update`.

To make sure we don't regress, I added a Postman test that fails without this patch and passes with this patch.


jjfreund:
LGTM, tested locally.

PR #204, opened 2017-10-07
Revive rake db:populate

richardxia:
This fixes `rake db:populate` and spruces up the schedule generator so that schedule are more random. Resources will now be generated with potentially multiple or no schedules per day which will hopefully catch more kinds of UI bugs. I've updated the README to suggest that you can use `rake db:populate` in place of `rake linksf:import` for fixture generation.

I also configured Travis to use `rake db:populate` instead of the LinkSF import for the Postman tests.


richardxia:
@jjfreund @lgarofalo ping.

lgarofalo:
LGTM, I'll merge once I verify it locally

lgarofalo:
I encountered this warning in CI and locally

`/home/travis/build/ShelterTechSF/askdarcel-api/vendor/bundle/ruby/2.3.0/gems/factory_girl-4.7.0/lib/factory_girl/evaluator.rb:68: warning: undefining `object_id' may cause serious problems`

richardxia:
> /home/travis/build/ShelterTechSF/askdarcel-api/vendor/bundle/ruby/2.3.0/gems/factory_girl-4.7.0/lib/factory_girl/evaluator.rb:68: warning: undefiningobject_id' may cause serious problems`

Yeah, that's actually due to the fact that we have a field on `ChangeRequest` named `object_id`, which overrides the Ruby builtin method on all objects named `object_id`. We should probably rename that column.

PR #226, opened 2017-10-11
Add JSON schema validation to Postman tests.

richardxia:
This adds JSON schema validation to some, but not all the Postman tests. The admin pages are completely unvalidated. I think these schemas would be useful to have outside of Postman as well, since it could serve as documentation or even be used in Rails or JS tests, but I'm not sure what's a good way of sharing the schema. I'm also not totally sure if my schema is completely correct for cases where we may return both null or empty strings, or null and empty arrays. It'd help to tighten up these semantics so that API clients could have better guarantees about the returned data structures.

I think these changes are worth merging before I add schemas for the rest of the Postman tests because I think there's a high risk of getting merge conflicts in the exported Postman JSON files.

### DIFF
--- a/postman/AskDarcel%20API.postman_collection.json
+++ b/postman/AskDarcel%20API.postman_collection.json
@@ -1,12 +1,253 @@
 {
-	"variables": [],
 	"info": {
 		"name": "AskDarcel API",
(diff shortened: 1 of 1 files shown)


richardxia:
@jjfreund @lgarofalo ping.

lgarofalo:
LGTM

PR #230, opened 2017-10-26
Allow CORS for testing

lgarofalo:


lgarofalo:
Tested locally, looking to try out in staging.

Issue #209, opened 2017-10-08
Admin analytics


jjfreund:
Create basic reports that convey number of resources and services added, change logs filed and approved, and number of resources/services currently active per category. Build a front end to display these reports.

twolfe2:
@jjfreund could we have 2 endpoints for this? 

One for the post workshop stats that would be passed this:
```
start_date: 2017-10-01
end_date: 2017-10-02
```

and return then return the following:
```
{
  change_requests: 221,
  resources_added: 10,
  services_added: 50,
  logs_filed: 300,
  logs_approved: 21
}
```

And another for the category stats that will return something like this:
```
[
  {
    name: "Housing",
    services: 200,
    resources: 43
  }, 
  {
    name: "Health"
    services: 210,
    resources: 34
  }
]
```

lgarofalo:
@twolfe2 I will start working on these endpoints

PR #231, opened 2017-10-29
feat(analytics): added admin analytics data

lgarofalo:
Example output
GET: /categories/counts
`[
    {
        "name": "Food",
        "services": 44,
        "resources": 44
    },
    {
        "name": "Hygiene",
        "services": 14,
        "resources": 14
    },
    {
        "name": "Medical",
        "services": 105,
        "resources": 105
    },
    {
        "name": "Shelter",
        "services": 24,
        "resources": 24
    },
    {
        "name": "Technology",
        "services": 55,
        "resources": 55
    }
]`

### DIFF
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -10,7 +10,6 @@ class CategoriesController < ApplicationController
     render json: CategoryPresenter.present(categories)
   end
 
-  # rubocop:disable Metrics/AbcSize
   def counts
     if !admin_signed_in?
       render status: :unauthorized
@@ -18,10 +17,9 @@ class CategoriesController < ApplicationController
       render status: :ok, json:
           Category.order(:name).map { |c|
             { name: c.name,
-              services: Service.joins(:categories).where('categories.id' => c.id).where('status' => 1).count,
-              resources: Resource.joins(:categories).where('categories.id' => c.id).where('status' => 1).count }
+              services: c.services.where('status' => 1).count,
+              resources: c.resources.where('status' => 1).count }
           }
     end
   end
-  # rubocop:enable Metrics/AbcSize
 end
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -9,4 +9,19 @@ class CategoriesController < ApplicationController
     categories = categories.where(top_level: top_level) unless top_level.nil?
     render json: CategoryPresenter.present(categories)
   end
+
+  # rubocop:disable Metrics/AbcSize
+  def counts
+    if !admin_signed_in?
+      render status: :unauthorized
+    else
+      render status: :ok, json:
+          Category.order(:name).map { |c|
+            { name: c.name,
+              services: Service.joins(:categories).where('categories.id' => c.id).where('status' => 1).count,
+              resources: Resource.joins(:categories).where('categories.id' => c.id).where('status' => 1).count }
+          }
+    end
+  end
+  # rubocop:enable Metrics/AbcSize
 end
--- a/app/controllers/change_requests_controller.rb
+++ b/app/controllers/change_requests_controller.rb
@@ -74,6 +74,23 @@ class ChangeRequestsController < ApplicationController
     end
   end
 
+  def pending_count
+    if !admin_signed_in?
+      render status: :unauthorized
+    else
+      render json: {
+                    address: ChangeRequest.all.where('type' => 'AddressChangeRequest').count,
+                    note: ChangeRequest.all.where('type' => 'NoteChangeRequest').count,
+                    phone: ChangeRequest.all.where('type' => 'PhoneChangeRequest').count,
+                    resource: ChangeRequest.all.where('type' => 'ResourceChangeRequest').count,
+                    schedule_day: ChangeRequest.all.where('type' => 'ScheduleDayChangeRequest').count,
+                    service: ChangeRequest.all.where('type' => 'ServiceChangeRequest').count,
+                    new_resources: Resource.all.where('status' => 0).count,
+                    new_services: Service.all.where('status' => 0).count
+                    }
+    end
+  end
+
   def replace_field_changes(change_request)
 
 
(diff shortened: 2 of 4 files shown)


app/controllers/categories_controller.rb
```diff
@@ -9,4 +9,19 @@ def index
...
     render json: CategoryPresenter.present(categories)
   end
+
+  # rubocop:disable Metrics/AbcSize
+  def counts
+    if !admin_signed_in?
+      render status: :unauthorized
+    else
+      render status: :ok, json:
+          Category.order(:name).map { |c|
```
richardxia: I think you want to use `do/end` instead of curly braces here. There's actually a difference between the two when it comes to syntax and binding strength: https://gemfile.wordpress.com/2014/04/04/do-end-vs-not-just-a-matter-of-style/

app/controllers/categories_controller.rb
```diff
@@ -9,4 +9,19 @@ def index
...
+            { name: c.name,
+              services: Service.joins(:categories).where('categories.id' => c.id).where('status' => 1).count,
+              resources: Resource.joins(:categories).where('categories.id' => c.id).where('status' => 1).count }
```
richardxia: You don't need to do the manual join here since we define the [many to many relationship](https://github.com/ShelterTechSF/askdarcel-api/blob/54b93b7fb279fd54f3471667ab932c4c3e53195a/app/models/category.rb#L2-L3) in the model. I think this can be written more succinctly like this:

```ruby
c.services.where(status: 1).count
```

app/controllers/categories_controller.rb
```diff
@@ -9,4 +9,19 @@ def index
     categories = categories.where(top_level: top_level) unless top_level.nil?
     render json: CategoryPresenter.present(categories)
   end
+
+  # rubocop:disable Metrics/AbcSize
+  def counts
+    if !admin_signed_in?
+      render status: :unauthorized
```
richardxia: We should consider refactoring this into a [`before_action`](http://guides.rubyonrails.org/action_controller_overview.html#filters), but probably in a later PR.

lgarofalo: Yea, good point.

app/controllers/categories_controller.rb
lgarofalo: Ah, ill test that.

lgarofalo: Nice, that works.

app/controllers/categories_controller.rb
lgarofalo: That actually didn't work, will require more reading, so i'll leave as is for now.

lexholden:
Ideally, for the current spec of the stats admin panel, we'd require some additional information in this api. Here's what I understood the admin stats page to contain:

- Summary of open change request counts without having to load all of them (changeRequestCount, fieldChangeCount, proposedServiceCount)
- Some method of measuring change requests approved over a given time period, to track success within a datathon
- Number of resources/services in total, and broken down by category. We can link through to these categories also.
- Longer term, some widgets embedded from Google Analytics so we can track visitors to AskDarcel in a single pane of glass

Unless there's any objections to creating an API to satisfy the above, I'd suggest the UI team create a dummy page assuming the above APIs are available with dummy data, and the APIs can be added in a future sprint.

lgarofalo:
@lexholden Yes, theres a couple other tickets in this epic for those, this is for 3) in your list, 

https://github.com/ShelterTechSF/askdarcel-api/issues/209 has tickets for the other Change request info.

Happy to make any changes we want to the resource/service counts, I was just tackling these one at a time.

In regard to tracking, I have started setting up some Google Analytics, let me follow up on that after this epic, https://www.sheltertech.org is already registered, maybe I can share access to that.

lgarofalo:
@lexholden For the format of all pending change requests
GET: /change_requests/pending_count 

{
    "address": 0,
    "note": 0,
    "phone": 0,
    "resource": 216,
    "schedule_day": 0,
    "service": 242,
    "new_resources": 0,
    "new_services": 0
}

lgarofalo:
For time frame

GET: /change_requests/activity_by_timeframe?start_date=2015-06-11T11:34+01:00&end_date=2019-06-11T11:34+01:00
{
    "new": {
        "address_cr": 0,
        "note_cr": 0,
        "phone_cr": 0,
        "resource_cr": 216,
        "schedule_day_cr": 0,
        "service_cr": 242,
        "new_resources": 0,
        "new_services": 0
    },
    "approved": {
        "address_cr": 0,
        "note_cr": 0,
        "phone_cr": 0,
        "resource_cr": 0,
        "schedule_day_cr": 0,
        "service_cr": 0,
        "new_resources": 0,
        "new_services": 0
    }
}

postman/AskDarcel Admin API.postman_collection
```diff
@@ -0,0 +1,550 @@
+{
```
jjfreund: I think you forgot to escape the spaces in the name, causing this to show up as a new file...

app/controllers/change_requests_controller.rb
```diff
@@ -74,6 +74,54 @@ def approve
...
+  def activity_by_timeframe
+    if !admin_signed_in?
+      render status: :unauthorized
+    else
+      start_date = params[:start_date].to_s
+      end_date = params[:end_date].to_s
+      render json: {
+          new: {
+                address_cr: ChangeRequest.all.where('type' => 'AddressChangeRequest').where("DATE(created_at) > ?", start_date.to_datetime).where("DATE(created_at) < ?", end_date.to_datetime).count,
+                note_cr: ChangeRequest.all.where('type' => 'NoteChangeRequest').where("DATE(created_at) > ?", start_date.to_datetime).where("DATE(created_at) < ?", end_date.to_datetime).count,
```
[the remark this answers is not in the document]
lgarofalo: Sure, will update

lgarofalo:
@jjfreund Fixed your comment, going to merge this for now so its available in staging.

PR #233, opened 2017-10-30
fix(auth): removed the token cycling configuration for admin auth

lgarofalo:
Only requirement is that you need to 

update admins set tokens = null;

Or you will run into a 500 internal server error on authing.


config/initializers/wrap_parameters.rb
```diff
@@ -5,7 +5,7 @@
 
 # Enable parameter wrapping for JSON. You can disable this by setting :format to an empty array.
 ActiveSupport.on_load(:action_controller) do
-  wrap_parameters format: [:json] if respond_to?(:wrap_parameters)
+  wrap_parameters format: [] if respond_to?(:wrap_parameters)
```
lgarofalo: This change was actually unnecessary, removing.

PR #241, opened 2017-12-03
238 and 239 inactivation endpoints

jjfreund:


app/controllers/resources_controller.rb
```diff
@@ -29,6 +29,16 @@ def create
     end
   end
 
+  def destroy
```
lgarofalo: I like the soft delete here.

lgarofalo:
LGTM

PR #242, opened 2017-12-03
chore(migrations): fix api image build

lgarofalo:
This auto-run migrations code that wasn't working wasn't properly removed.


lgarofalo:
cc @jjfreund

PR #249, opened 2017-12-07
Pin version of newman.

richardxia:
The latest version of newman (3.9.0) seems to be causing issues with variable substitution not working. Pinning 3.8.3 until we resolve this issue.


lgarofalo:
LGTM

PR #252, opened 2017-12-14
Fix Postman schema validation.

richardxia:
The old API for setting global variables doesn't seem to work anymore (`postman.setGlobalVariable()`). This silently broke schema validations because attempting to get an undefined global variable would return `undefined`, and apparently when `undefined` is passed into the schema validator, it just passes validation. This was fixed by using the new API for setting global variables (`pm.globals.set()`).

I also got rid of the dummy "Set global variables" test and replaced it with the new [collection-wide pre-request scripts](https://www.getpostman.com/docs/postman/scripts/intro_to_scripts#execution-order-of-scripts) that get run before every test, which was added in the newest release of Postman 4 days ago.

One last thing I did was add a specific check to make sure that the schema that we try to retrieve from a global variable is not null or undefined, so if this happens again in the future, the tests will actually fail.


richardxia:
Apparently the old version of newman doesn't work with my changes, so this is blocked by #251. I can confirm that the latest version of newman *does* pass locally.

PR #254, opened 2017-12-18
Certification Endpoints

lexholden:


postman/AskDarcel%20Admin%20API.postman_collection.json
```diff
@@ -1,9 +1,10 @@
 {
+	"variables": [],
 	"info": {
 		"name": "AskDarcel Admin API",
-		"_postman_id": "a19cf529-de95-3738-10f3-2568497f9084",
+		"_postman_id": "241f6b3d-c648-de5c-61a0-6d4c5a7bfe58",
 		"description": "",
-		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
+		"schema": "https://schema.getpostman.com/json/collection/v2.0.0/collection.json"
```
richardxia: Can you export the v2.1.0 schema version of the collection? I think that's why the diff is so large.

config/routes.rb
```diff
@@ -13,6 +13,8 @@
     end
 
     post :create
+    delete :inactviate
```
richardxia: Is this thing used at all? If not, then let's delete it. If it is, then let's fix the typo (`v` and `i` are in the wrong places), but even better, can we use `deactivate` as the verb instead, which is think is more [grammatically correct](https://english.stackexchange.com/questions/312339/inactive-or-deactivated/312346).

postman/AskDarcel%20Admin%20API.postman_collection.json
```diff
@@ -308,7 +313,60 @@
...
+						"key": "uid",
+						"value": "dev-admin@sheltertech.org"
+					}
+				],
+				"body": {
+					"mode": "raw",
+					"raw": ""
+				},
+				"url": {
+					"raw": "{{base_url}}/services/1/approve",
```
richardxia: Shouldn't this be `/services/1/certify`?

jjfreund:
Made the changes requested, thanks. As far as your question goes: this endpoint is currently available to anonymous visitors, but so are other edit functions.

Issue #261, opened 2018-01-18
Configure separate Algolia indices for different environments


richardxia:
https://github.com/algolia/algoliasearch-rails#per-environment-indices

This will let us use the same Algolia application on both prod and staging but not have their indexes interfere with each other.

PR #267, opened 2018-02-05
Add ALGOLIA_INDEX_PREFIX environment variable.

richardxia:
This is a slightly different approach than what was originally proposed in https://github.com/sheltertechsf/askdarcel-api/issues/261, since the `per_environment` option just adds a `_#{Rails.env}` suffix to the indexes, and we actually use the same RAILS_ENV in both prod and staging.

Instead, I added an `ALGOLIA_INDEX_PREFIX` environment variable which is used to manually prepend a prefix to Algolia indexes. Note that we need to remember to do this with all future indexes that we create. Today, we only have a single index, so there's only one place in the code to add it, but if we add new indexes, then we need to add them to the other places.

I tested this by running it locally against our Algolia instance.

### DIFF
--- a/app/models/resource.rb
+++ b/app/models/resource.rb
@@ -23,7 +23,9 @@ class Resource < ActiveRecord::Base
   end
 
   if Rails.configuration.x.algolia.enabled
-    algoliasearch per_environment: true do
+    # Note: We can't use the per_environment option because both our production
+    # and staging servers use the same RAILS_ENV.
+    algoliasearch index_name: "#{Rails.configuration.x.algolia.index_prefix}_Resource" do
       geoloc :address_latitude, :address_longitude
 
       add_attribute :address do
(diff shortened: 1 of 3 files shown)


richardxia:
Also, I forgot to mention, we'll need to set the `ALGOLIA_INDEX_PREFIX` environment variable on both staging and prod.

richardxia:
Bump, for review.

richardxia:
> Looks like we will need to set ALGOLIA_INDEX_PREFIX in the environment templates as we deploy this and if its not set then the Algolia functionality will be disabled.

Yes, that's correct. Let me know if you prefer a different way of doing it. I chose this way because it would ensure that search still functions even if the Algolia settings are not all there.

> Is there any access / administrative information we should share about using different indices and algolia in general?

I think the only thing that is important is that we only use the prod index in production. We could set up separate user accounts for prod and staging to prevent accidentally using the wrong indexes, but I didn't want to overcomplicate things.

PR #272, opened 2018-02-18
Use globally-scoped docker-compose network.

richardxia:
This allows other docker containers to connect to the network created by this docker-compose instance, e.g. an askdarcel-web instance running in a different docker-compose instance.


richardxia:
Thanks for reviewing!

PR #273, opened 2018-02-18
263 open referral schema changes

jjfreund:


app/controllers/services_controller.rb
```diff
@@ -121,6 +122,10 @@ def transform_service_params!(service, resource_id)
       schedule = service[:schedule_attributes] = service.delete(:schedule)
       schedule[:schedule_days_attributes] = schedule.delete(:schedule_days) if schedule.key? :schedule_days
     end
+
+    puts 'hi'
```
lgarofalo: probably should remove these 'puts'

app/presenters/programs_presenter.rb
```diff
@@ -0,0 +1,12 @@
+class ProgramsPresenter < Jsonite
+	property :id
```
richardxia: Boo to using tabs.

db/migrate/20180201040240_create_join_table_service_address.rb
```diff
@@ -0,0 +1,8 @@
+class CreateJoinTableServiceAddress < ActiveRecord::Migration[5.0]
+  def change
+    create_join_table :services, :addresses do |t|
+      # t.index [:service_id, :address_id]
+      # t.index [:address_id, :service_id]
```
richardxia: Why is this commented out?

PR #289, opened 2018-04-03
askdarcel-web issue 414: create test data in db

zjipsen:
necessary for tests added here https://github.com/ShelterTechSF/askdarcel-web/pull/457 to pass in TravisCI

### DIFF
--- a/lib/tasks/populate.rake
+++ b/lib/tasks/populate.rake
@@ -277,7 +277,7 @@ namespace :db do
       test_category_names.each { |name| FactoryGirl.create(:category, name: name, id: 234) }
       Category.find_by_name!('Test_Category_Top_Level').update(top_level: true)
       test_categories = Category.where(name: test_category_names)
-      
+
       resource = FactoryGirl.create(:resource,
                                     name: "A Test Resource",
                                     short_description: "I am a short description of a resource.",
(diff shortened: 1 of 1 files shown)


lib/tasks/populate.rake
```diff
@@ -271,5 +271,33 @@ namespace :db do
 
       resource.services = services
     end
+
+    1.times do
```
richardxia: To make this more deterministic, maybe we should move this to the beginning and force it to be ID 1? I'm not exactly sure what Rails will do if you randomly create a record with a higher ID number, and whether that ID may collide if it still tries counting from the last assigned ID.

zjipsen: I originally had it at the beginning, but ran into an issue. For some reason if this test data is created before the rest of the data, FactoryGirl creates additional resources and assigns them to the test category. This is annoying because it means that the testcafe test has to select a specific resource from the listings page rather than just grabbing the first one, and there are other minor annoyances introduced. 
I can certainly fix this though. So far when populating the database with this file, Rails has not had a problem with ID collision. Lmk if I should go ahead and fix it. Thanks!

richardxia: I see, that makes sense. Then I think this LGTM to merge for now. Thanks for adding this!

PR #302, opened 2018-05-23
Added protection for blank lat/lng

twolfe2:


app/models/service.rb
```diff
@@ -30,9 +30,11 @@ class Service < ActiveRecord::Base
       add_attribute :_geoloc do
         if addresses.any?
           addresses.map do |a|
-            { lat: a.address_latitude, lng: a.address_longitude }
+            if a.address_latitude.present? & a.address_longitude.present?
```
richardxia: In Ruby, the single `&` operator is the bit-wise AND operator. You really should be using `&&`.

richardxia: Also, Rubocop doesn't like this style. Can you please fix it so that people branching from master don't unconditionally fail Rubocop? In general, we should at least wait for CI to pass before merging, unless if we have a strong reason to not wait.

config/application.rb
```diff
@@ -13,10 +13,10 @@ class Application < Rails::Application
     config.api_only = true
 
     # Algolia
-    config.x.algolia.application_id = ENV['ALGOLIA_APPLICATION_ID']
-    config.x.algolia.api_key = ENV['ALGOLIA_API_KEY']
+    config.x.algolia.application_id = ENV['ALGOLIA_APPLICATION_ID'] || 'J8TVT53HPZ'
+    config.x.algolia.api_key = ENV['ALGOLIA_API_KEY'] || 'f8a0543e1a8db699f6359eeb11ebcd9d'
```
richardxia: Should this key really be committed to an open-source repo like this?

PR #309, opened 2018-05-27
Algolia search config instructions

twolfe2:


README.md
```diff
@@ -33,6 +33,12 @@ $ docker-compose up api
 $ docker-compose stop
 ```
 
+Set up the Algolia search index
+  - Create a .env file in the root of your askdarcel-api directory
+  - Populate that file with the backend environment variables found [here](https://sheltertech.quip.com/2ft5Ax19Kc6h)
+  - Run `Resource.reindex!` and `Service.reindex!` in the `rails console`
```
alexanderturinske: I take it this is for bare metal? I assumed it was and so I did the following:
1. Pulled down the changes
2. Set up .env file
2. Repulled docker image, rebuilt container
3. Found docker container name/id: `docker ps`
4. Navigated inside the container: docker exec -it <CONTAINER_ID/NAME> bash
5. Opened the rails console: `bin/rains console`
6. Ran the rails commands: `Resource.reindex!; Service.reindex!`
Even this did not work though. I don't know if it contributed to helping or not. I ended up having to update the environment variables in web and force it to use my GH handle in config.example. It seems like our webpack isn't grabbing process.env? Do you know anything about this?

richardxia: The Rails side (both with and without Docker) should automatically be picking up the environment variables in the .env file. What exactly was failing for you?

I don't know what the setup instructions are for the frontend, but these instructions don't appear to cover that side.

twolfe2: @alexanderturinske Could you ensure the web side is configured correctly? Just updated the readme with instructions for algolia: https://github.com/ShelterTechSF/askdarcel-web/blob/master/README.md#algolia

alexanderturinske: Sorry for the late response.

Yeah, I was just trying things, which may or may not have been correct. I guess my thoughts on the part of running something in the rail console was that if one was running docker, they would need to run the rail commands in the console on the docker container, but I didn't realize that one could run commands in the rail console on their computer and docker would pick it up.

Yeah, I'll look at the web side.

PR #332, opened 2018-07-22
Updated mohcd funding logic

twolfe2:
Made it so our logic to determine if a service/resource is MOHCD funded will work with the new name of the category


twolfe2:
❗️ Note: `rake algolia:reindex` should be run in the rails console after this is released so that the algolia index gets updated with these latest changes ❗️

PR #353, opened 2018-11-11
chore(gems): update versions

lgarofalo:


lgarofalo:
Not quite sure what broke here, will have to dig.

PR #356, opened 2018-11-13
Fixed algolia index to add resource schedules to service

lexholden:


app/models/service.rb
```diff
@@ -77,6 +77,14 @@ class Service < ActiveRecord::Base
         end
       end
 
+      add_attribute :resource_schedule do
+        if resource.schedule.present?
+          resource.schedule.schedule_days.map do |s|
```
jjfreund: I don't think this logic is exactly right -- the resource schedule is also used if the service schedule is not present but has no days associated with it. See the use_resource_schedule method.

Issue #337, opened 2018-09-30
Improve error handling for change request failures 


jjfreund:
Currently, the API to submit change requests only returns a 500 upon failure. Return a more descriptive error message in the case of failure.

PR #363, opened 2018-11-18
Improve change request error handling (#337)

cliffcrosland:
Before, when a client submitted a change request containing an invalid
field, the API would return a 500 Internal Server Error status code, and
invalid ChangeRequest and FieldChange records would remain in the
database afterward.

Now, when a client submits a change request containing an invalid field,
the API returns a 400 Bad Request status along with an error message
describing the incorrect field. Additionally, the entire operation is
wrapped in a Postgres transaction, so invalid ChangeRequest and
FieldChange records inserts are rolled back when validation errors
occur.


lgarofalo:
Looks good! Thanks!

PR #365, opened 2018-11-26
Add open_times to list of attributes for faceting.

richardxia:
Part of the reason that the Open Now filter doesn't work on the frontend is because the `open_times` attribute needs to be added as an ["attribute for faceting"](https://www.algolia.com/doc/guides/searching/filtering/?language=javascript#how-to-format-your-data-to-be-used-as-filters), which is apparently something that is mandatory for filtering on string fields.

I still need to make some frontend changes, but this can go in sooner. We'll want to do a full reindex after this is merged in.


richardxia:
> I still need to make some frontend changes, but this can go in sooner. We'll want to do a full reindex after this is merged in.

Never mind about this. The frontend works perfectly fine as-is.

We'll probably want to do some refactoring of how the search parameters work in the frontend code, since it's a little bit confusing that we do the filtering by literally modifying the URL rather than capturing the filters in structured data, but at least the filtering seems to work very well right now.

PR #366, opened 2018-11-27
Add scheduled job for Algolia reindexing

cliffcrosland:
Use `whenever` gem to schedule an Algolia reindex job every hour.

Use `delayed_job` gem to run a worker to continuously process background
jobs, including this Algolia reindex job.

### DIFF
--- a/app/jobs/algolia_reindex_job.rb
+++ b/app/jobs/algolia_reindex_job.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+# Background job to reindex all relevant database records in Algolia.
+#
+# Usage:
+#
+# To send job to a background worker to execute:
+#
+# > AlgoliaReindexJob.perform_async
+#
+# To execute in-line immediately from Rails console:
+#
+# > AlgoliaReindexJob.new.perform
+#
+class AlgoliaReindexJob
+  def self.perform_async
+    Delayed::Job.enqueue(AlgoliaReindexJob.new)
+  end
+
+  def perform
+    Resource.where(status: :approved).reindex
+    # Since Service and Resource share an algolia index, we use `reindex!` so
+    # Service reindexing does not clear out Resource index records.
+    Service.where(status: :approved).reindex!
+  end
+end
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,28 @@
 version: "3.5"
+
+# Common configuration for Rails processes.
+x-rails-common:
+  &rails-common
+  build:
+    context: .
+    dockerfile: Dockerfile.dev
+  depends_on:
+    - db
+  environment:
+    DATABASE_URL: postgres://postgres@db/askdarcel_development
+    TEST_DATABASE_URL: postgres://postgres@db/askdarcel_test
+    ALGOLIA_APPLICATION_ID:
+    ALGOLIA_API_KEY:
+    ALGOLIA_INDEX_PREFIX:
+  networks:
+    - askdarcel
+  volumes:
+    - .:/usr/src/app
+    # Avoid leftover server.pid files when container exits.
+    # https://auth0.com/blog/ruby-on-rails-killer-workflow-with-docker-part-1/
+    - type: tmpfs
+      target: /usr/src/app/tmp/pids
+
 services:
   db:
     image: "postgres:9.5"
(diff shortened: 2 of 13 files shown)


cliffcrosland:
From what I can tell, the CodeFactor Rubocop tests are flagging `db/schema.rb` and the new `db/migrate/*_create_delayed_jobs.rb` file. Since these are auto-generated ruby files, perhaps we can ignore the warnings?

lgarofalo:
This is a good idea, answers to some questions

* Yea let's ignore that CodeFactor stuff here, can dig into that later.
* It sounds like we can bake something into the image creation to handle the cron, might need to play around with that but it should all be available in this repo.

docker-compose.yml
```diff
@@ -41,6 +41,28 @@ services:
     volumes:
       - .:/usr/src/app
 
+  worker:
```
lgarofalo: Since this is just based off the same Dockerfile anyway, is it possible to combine this into the main api container?

lgarofalo:
cc @jjfreund and @richardxia for SME on the Algolia setup.

docker-compose.yml
richardxia: Is there a benefit to combining it with the main API container? I thought part of the point of Docker was so that it's really easy to create separate containers for each process.

If you're just talking about the repetition of the configuration between the `api` and `worker` services in docker-compose.yml, rather than the actual run-time of the two processes, we could make use of YAML's syntax to define a hashmap/object with the common attributes and include them into both the `api` and `worker` service. See the examples in https://docs.docker.com/compose/compose-file/#extension-fields

app/jobs/algolia_reindex_job.rb
```diff
@@ -0,0 +1,24 @@
...
+# > AlgoliaReindexJob.new.perform
+#
+class AlgoliaReindexJob
+  def self.perform_async
+    Delayed::Job.enqueue(AlgoliaReindexJob.new)
+  end
+
+  def perform
+    Resource.reindex
+    Service.reindex
```
richardxia: This is going to logically conflict with @jjfreund's changes in https://github.com/ShelterTechSF/askdarcel-api/pull/367, where he's slightly modifying the reindexing commands.

I think we should pull the logic into a common place where both the Rake task and this delayed job call the same function.

docker-compose.yml
lgarofalo: Yea, in this instance I feel running/managing an additional container to run a cron is under the threshold of value for the extra overheard. It's a balance, but in this case I would lean towards simplicity.

app/jobs/algolia_reindex_job.rb
lgarofalo: His change is only to reindex approved services, right? I don't see a conflict there.

docker-compose.yml
cliffcrosland: @richardxia Sounds great. I agree - I think it might be best to keep the `api` and `worker` containers separate like they currently are in this PR after all. Nevertheless, given my unfamiliarity with the GCP devops infrastructure, I'm also open to combining them if it's needed.

Now that the `docker-compose.yml` file has been refactored, is it ok to merge? I'll spend some time looking at the gcloud repo to see if I can figure out which changes are needed in the deployment setup.

app/jobs/algolia_reindex_job.rb
jjfreund: Richard is right, Cliff should use the same business logic I do (I believe). There are two things that seem wrong here -- first, the where clauses do not restrict the results to active statuses (see my commit for reference). Second, since the services and resources share an index, the second statement has to use the reindex! method, or else it will blow away the records created by the first index.

cliffcrosland:
See also https://github.com/ShelterTechSF/gcloud/pull/3

app/jobs/algolia_reindex_job.rb
cliffcrosland: @jjfreund Thanks! I've pushed a change to make it so that the AlgoliaReindexJob uses the following commands. I also changed the `algolia.rake` script to use this common code. Any thoughts?
```Ruby
    Resource.where(status: :approved).reindex
    # Since Service and Resource share an algolia index, we use `reindex!` so
    # Service reindexing does not clear out Resource index records.
    Service.where(status: :approved).reindex!
```

PR #367, opened 2018-11-28
350 fix resource and service inactivation

jjfreund:


lib/tasks/algolia.rake
```diff
@@ -3,8 +3,8 @@
 namespace :algolia do
   task reindex: :environment do
     print "Reindexing resource/service index... "
-    Resource.reindex!
-    Service.reindex!
+    Resource.where(status: :approved).reindex
```
lgarofalo: With the `.where`, can this become null? (looking at testerrors)

richardxia: Looking at the Travis errors, it looks like it's actually erroring on the `remove_from_index!` calls in the other lines of the diff.

lgarofalo: Ah ok, I'll take a deeper look.

lgarofalo:
@jjfreund good to merge?

PR #380, opened 2019-01-22
Introduce new system for Kubernetes CronJobs

cliffcrosland:
Traditional cronjobs do not work with Kubernetes. Introduce the
`cron-kubernetes` gem to manage cronjobs for our Kubernetes cluster.

Right now, when a deployment to staging occurs, `bin/worker_restart.sh` is executed. This script will update Kubernetes CronJobs and restart the `delayed_jobs` worker.


lgarofalo:
Given that both the existing api and scheduled job k8s configurations use the same image, how is the different role assigned to both at runtime?

config/k8s-cron-template.yml
```diff
@@ -0,0 +1,13 @@
...
+kind: CronJob
+metadata:
+  name: scheduled-job
+spec:
+  template:
+    metadata:
+      name: scheduled-job
+    spec:
+      containers:
+      - name: index.docker.io/sheltertechsf/askdarcel-api:latest
```
richardxia: I have experience with Kubernetes but not Google Kubernetes Engine; is there a way to ensure that the image tag is consistent between the api pods and this job? My fear here is that the way this is written, any time someone merges to master, we'll build a new image and push it to `latest`, and in production we'll start running whatever the latest image is, not the latest *deployed* image.

I'm more familiar with using [Helm](https://helm.sh/) to parameterize your Kubernetes YAML files and ensure that the image tags are the same across multiple pods that are part of the same "release". For example, taking this open source Grafana Helm Chart, you can specify the image in one place in [values.yaml](https://github.com/helm/charts/blob/bb5b5ab85caf47966d02ed24ec64f6a751b46956/stable/grafana/values.yaml#L290) while referring to it in multiple locations [[1]](https://github.com/helm/charts/blob/bb5b5ab85caf47966d02ed24ec64f6a751b46956/stable/grafana/templates/deployment.yaml#L92) [[2]](https://github.com/helm/charts/blob/bb5b5ab85caf47966d02ed24ec64f6a751b46956/stable/grafana/templates/deployment.yaml#L111).

cliffcrosland:
@lgarofalo Great question. Given that I'm a k8s newbie, this might be incorrect, but here is what I understand from the k8s documentation about [Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/) and [CronJobs](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/). 

It appears that each scheduled CronJob will do the following:
- Spin up a pod that matches the YML definition in `k8s-cron-template.yml`.
- As defined in the YML template, this pod simply consists of one container: `api`.
- When the pod spins up, the `api` container will run the command associated with this `CronJob`. eg. `bundle exec rails runner ''AlgoliaReindexJob.perform_async"`.
- When the command finishes, the pod spins down.

That said, it seems strange to me that k8s spins up a full pod and then tears it back down every time a job runs. Perhaps I'm wrong about how it works 😕 

If each CronJob spins up a new pod, then the main pod running the rails server should remain isolated and safe from interference (I think).

richardxia:
> That said, it seems strange to me that k8s spins up a full pod and then tears it back down every time a job runs. Perhaps I'm wrong about how it works 😕

I have a decent amount of K8s experience, and although I haven't used CronJobs before, this sounds right based on how Jobs work. I think part of the idea is that it's cheap to spin up a Pod, unlike spinning up an entire VM, so it isn't actually that great of a cost to do it for each run.

cliffcrosland:
Closed and re-opened to get Travis CI to rebuild. For some reason, it didn't re-build after I pushed a recent commit to fix a Rubocop warning.

cliffcrosland:
I think I have a better system for this all-together. I think it would be best to keep all k8s configuration in the ShelterTechSF/gcloud repo. Whenever we want to add a CronJob, we can add it to the k8s yaml files there.

Closing this PR in favor of this new PR: https://github.com/ShelterTechSF/gcloud/pull/7

Issue #383, opened 2019-02-10
Correct featured categories in dev database, and add featured services for a specific category (Legal)


jjfreund:
Modify the data setup scripts such that the featured categories reflect the new landing page spec (Basic Needs & Shelter, Eviction Prevention, Affordable Housing, Health & Medical, Employment, Legal) and make some features in the legal category featured. Currently developers are working locally with a data set not representative of the production data set.

PR #385, opened 2019-02-10
Correct featured categories in dev DB (#383)

cliffcrosland:


lib/sheltertech/db.rb
```diff
@@ -379,7 +407,12 @@ module Constants # rubocop:disable Metrics/ModuleLength
         'Re-entry Services',
         'Clean Slate',
         'Probation and Parole',
-        'MOHCD Funded Services'
+        'MOHCD Funded Services',
+        'Basic Needs & Shelter',
+        'Eviction Prevention',
```
jjfreund: You can probably remove Eviction Defense, the category Eviction Prevention replaces. Also according to the requirements doc we are going to go with Housing instead of Affordable Housing:

https://sheltertech.quip.com/K0ZMA1vgbxjd/SFServiceGuideorg-Landing-Page-Requirements

PR #397, opened 2019-03-26
adds presence validation for open and close hours

tsyaeger:
Requires opens_at and closes_at fields for ScheduleDay creation.


tsyaeger:
Joe, this adds a delete action to the change request controller persist method and a basic validation test spec. I've moved the logic for that one action type to the model. Would be happy to refactor the entire method, as the controller could stand to be a little leaner.

lgarofalo:
LGTM

Issue #399, opened 2019-03-31
Store open and close times in schedule_day table as times


jjfreund:
Currently, the open and close time are stored as integers, but could be more accurately represented as as times. There is a postgres data type, time, which is used to represent time-of-day (dateless) values. Modify the schedule_day table to use this data type for open and close times.

The day of week is currently a single field in the schedule_day table. This is incorrect -- since open and close times can span one day to another, we should track the open day and close day separately. Add the following four fields to the schedule day table:

open_day character
open_time time
close_day character
close_time time

Maxastuart:
This is perfect! Thanks so much!

PR #400, opened 2019-03-31
399 improve open and close times

jjfreund:
See #399 -- Added time fields for more intuitive storage of open and close times, and added open and close day fields to represent cases where open and close occur across days. Left old fields in for backwards compatibility; will remove when front end and external partners are using new fields.


lgarofalo:
@jjfreund You got a conflict here to fix

PR #411, opened 2019-04-21
Enable eager loading of code in order to avoid deadlocking issue.

richardxia:
I found a GitHub ticket on the Puma project that described a problem similar to what we were seeing, and there's a workaround that they suggested that seemed to work for me: https://github.com/puma/puma/issues/1184

This PR turns on `eager_load` in both development and testing environments, which apparently works around the deadlocking issue that can cause the API server to hang.


tsyaeger:
This has prevented the deadlocking for me.

AnnaKalkanis:
Ya, this looks good!

PR #420, opened 2019-05-08
395 Use rspec-rails-swagger and write trivial integration tests

kumquatexpress:
rspec-rails-swagger hooks into existing request specs and uses their responses to generate swagger JSON. The output is written to the same `swagger/v1/swagger.json` file as before, and served at `/api-docs/index.html`.

Upside is that writing test specs for API endpoints automatically outputs formatted documentation -- the downside is that there's very little coverage at the moment, but writing trivial test cases for generating docs is pretty easy.

Attached is a dummy image of the generated docs from the `/services/:id` route spec.
example


kumquatexpress:
@richardxia wasn't sure who to tag and you were the closest POC on the ticket -- feel free to add someone else as reviewer!

richardxia:
I've never used Swagger before, but these changes LGTM at a high level. I tagged a bunch of other people who work on the backend to review this PR.

BTW, you are failing an RSpec test in Travis.

lgarofalo:
We do some URL rewriting with NGINX at the deployed environments, I can help make this available in production but right now the expected paths would fail.

kumquatexpress:
The real benefit here is for faster iteration in development by having reliable API docs/playground environment. We can productionize later on to make it more accessible though!

PR #426, opened 2019-06-13
Introduce Algolia incremental index job

cliffcrosland:
This job scans through all of the resources that have been updated since
the last time it ran. Approved resources and their approved services are
updated in Algolia, and all others are removed from Algolia.

We introduce a new table named `bookmarks` that allows the job to pick
up again from where it left off the last time it ran.

Updated ChangeRequestsController to make it so the `updated_at` time of
the associated Resource is touched whenever a ChangeRequest is applied.
Allows the Algolia incremental index job to detect updates.

Remove the deprecated `Delayed::Job` system. It has been replaced for a
while now by the k8s CronJob setup.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,11 +30,6 @@ gem 'lograge'
 
 gem 'prometheus-client'
 
-# Background Job Processing
-gem 'daemons'
-gem 'delayed_job_active_record'
-gem 'whenever', require: false
-
 gem 'rspec-rails'
 
 group :development, :test do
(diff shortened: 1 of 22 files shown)


cliffcrosland:
Hmm thought about this some more. There are records related to resources and services in other tables (eg. addresses, schedules) that are indexed in Algolia but can be updated without the updated_at time of the associated resource or service being touched. 

We would miss such updates in this job. I have a few more ideas to implement before this is ready

cliffcrosland:
Cool. This should work great now. Simplified it a bit too. Here's how it works:

- In the ChangeRequestsController, we touch the `updated_at` time of the Resource. This signals to the Algolia incremental index job that a Resource has been updated.
- The Algolia incremental index job now just scans through updated Resources. It will sync each updated Resource and all of its Services to Algolia.

PR #431, opened 2019-06-30
422 deactivate services when parent resource is deactivated

jjfreund:


app/services/resources/resources.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Services
+  module Resources
+    include Errors
```
cliffcrosland: Rails might auto-load this module, so I think it might be possible to delete this `include Errors` line here. In development mode, Rails lazily loads modules, so when it encounters `raise Errors::PreconditionFailed`, it will go look for the appropriate file and load it into scope.

app/services/services/services.rb
```diff
@@ -0,0 +1,20 @@
...
+    include Errors
+
+    def self.deactivate(id)
+      service = Service.find id
+      raise Errors::PreconditionFailed unless service.approved?
+      service.inactive!
+      remove_from_algolia(service)
+    end
+
+    def self.remove_from_algolia(service)
```
cliffcrosland: Maybe this can be factored out as a common utility since it is used in both services? Might be able to put it in `app/services/utils.rb`? Perhaps the error log message can be generalized as follows:

```ruby
def self.remove_from_algolia(record)
  record.remove_from_index!
rescue StandardError
  puts "failed to remove #{record.class} #{record.id} from algolia index"
end
```

app/services/resources/resources.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Services
+  module Resources
```
cliffcrosland: Might be cool to add some rspec tests for these services, but if they are already covered by the Postman tests, that's cool.

cliffcrosland:
Haha, looks like Rubocop wants you to add empty newlines after the `Errors::PreconditionFailed` guard clauses

Issue #402, opened 2019-04-04
Setup Sentry Environments


Maxastuart:
- Set up environments in Sentry to filter the errors from the Production server and the Staging server (and none from local machine instances) 
- Filter slack notifications (#sentry-alerts) to only show Production errors

expected behavior: errors from Prod get reported to sentry prod environment and #sentry-alerts slack channel. errors from Staging get reported to sentry prod environment. errors from local machines do not get reported to Sentry.

actual behavior: while running the askdarcel app anywhere (even on local machines) all errors are reported to Sentry (which also posts to #sentry-alerts in Slack). 

how to set it up with sentry:
https://docs.sentry.io/enriching-error-data/environments/?platform=javascript

PR #432, opened 2019-07-01
402: Sending environment in config to Sentry

rtindru:
Giving ability to filter error reports on Sentry by sending environment
information during initialization
https://docs.sentry.io/enriching-error-data/environments/?platform=javascript

Issue: https://github.com/ShelterTechSF/askdarcel-api/issues/402
Branch: 402-sentry-env-split


config/initializers/sentry.rb
```diff
@@ -4,7 +4,8 @@
   SENTRY_PUBLIC_KEY = 'd35c8b2336084e67b08f7c4b5eeb54b9'
   SENTRY_PROJECT_ID = '1291517'
   config.dsn = "https://#{SENTRY_PUBLIC_KEY}@sentry.io/#{SENTRY_PROJECT_ID}"
-
+  # Send the environment to Sentry (development/testing/production)
+  config.environments = [Rails.env]
```
richardxia: I'm not sure if this will do example what we want. I believe that in the staging site, the `Rails.env` is still set to `production`.

PR #462, opened 2019-10-17
Enable a resource to have multiple addresses

alexanderturinske:
Resources Routes modified
- [x] GET /resources/:id
- [x] POST /resources/new
- [x] POST /resources/:id/change_requests
    - it seems that when an edit happens to the address, the UI makes to requests
    - first to /resources/:id/change_requests with the payload { "change_request": {} }
    - second to /addresses/250/change_requests with the payload containing the actual changes
        - example: {"address_change_request":{"id":22230,"status":"pending","type":"AddressChangeRequest","object_id":250,"field_changes":[{"field_name":"name","field_value":"National Center for Youth Law (NCYL)"}]}}
    so this will support multiple addresses because the UI will make multiple change request POSTs, but the UI logic might have to be updated to account for an array of addresses
- [x] DELETE /resources/:id
    - this is handled by rails and updating the relationship

Additionally resources changes made:
- [x] update address to addresses where necessary
- [x] Services has a resource associated with them. That resource should have the same addresses property as when you GET just the resource itself.
- [x] Algolia returns resources and services as hits. These hits should have an addresses property with an array of addresses
- [x] The algolia hits have _geoloc property with lat and long data for the resource. Now that there are multiple locations for any resource, we should move these lat/long properties onto the address they are associated with in the addresses array.
- [x] Reindexing of algolia runs without errors and searches return new format

multiple-addresses

From https://github.com/ShelterTechSF/askdarcel-web/pull/894


alexanderturinske:
This change only works for `GET` requests; more work needs to be done for the `POST` request and it is in progress

app/models/resource.rb
```diff
@@ -11,7 +11,7 @@ class Resource < ActiveRecord::Base
 
   has_and_belongs_to_many :categories
   has_and_belongs_to_many :keywords
-  has_one :address, dependent: :destroy
+  has_many :address, dependent: :destroy
```
alexanderturinske: should probably update this to be plural

postman/AskDarcel%20API.postman_collection.json
```diff
@@ -1127,7 +1127,51 @@
...
+					],
+					"path": [
+						"resources"
+					]
+				}
+			},
+			"response": []
+		},
+		{
+			"name": "Submit New Resource with multiple addresses",
```
alexanderturinske: the diff is weird here, but I update the payload above to wrap the address in `[...]` and added this test for multiple addresses

AnnaKalkanis:
Ya, I think it would be nice if this were a new pluralized property `addresses`

swagger/v1/swagger.json
```diff
@@ -4,53 +4,1552 @@
     "title": "sheltertech API",
     "version": "v1"
   },
+  "definitions": {
```
jjfreund: Why did these changes occur?

alexanderturinske: I don't know how these changes happened, so I am deleting them.

alexanderturinske: I don't know how these changes happened, so I am deleting them. Good catch

alexanderturinske: Done

app/models/resource.rb
Maxastuart: I believe this has to be updated via a migration (`rails generate migration ...`) rather than manually changing the file

alexanderturinske: @Maxastuart researched this together and confirmed it does not need a migration.

app/models/service.rb
```diff
@@ -33,16 +33,6 @@ class Service < ActiveRecord::Base
 
       # Define attributes used to build an Algolia record
       add_attribute :status
-      add_attribute :_geoloc do
```
alexanderturinske: moved onto `:addresses` below

app/models/service.rb
```diff
@@ -52,17 +42,23 @@ class Service < ActiveRecord::Base
               state_province: a.state_province,
               postal_code: a.postal_code,
               country: a.country,
-              address_1: a.address_1
+              address_1: a.address_1,
+              latitude: a.latitude.to_f || nil,
+              longitude: a.longitude.to_f || nil,
```
alexanderturinske: added to match what the UI expects in https://github.com/ShelterTechSF/askdarcel-web/pull/894/

app/models/service.rb
```diff
@@ -52,17 +42,23 @@ class Service < ActiveRecord::Base
               state_province: a.state_province,
               postal_code: a.postal_code,
               country: a.country,
-              address_1: a.address_1
+              address_1: a.address_1,
+              latitude: a.latitude.to_f || nil,
```
jjfreund: Why did you add this check in?

app/models/resource.rb
```diff
@@ -38,23 +38,24 @@ class Resource < ActiveRecord::Base
       # specify the list of attributes available for faceting
       attributesForFaceting %i[categories open_times]
       # Define attributes used to build an Algolia record
-      add_attribute :_geoloc do
-        { lat: address_latitude.to_f, lng: address_longitude.to_f }
```
jjfreund: Is removing this going to prevent Algolia from properly ordering by distance?

alexanderturinske: Ah, [you are correct](https://www.algolia.com/doc/guides/managing-results/refine-results/geolocation/). Adding back.

app/models/service.rb
alexanderturinske: adding back after Joe's comment above

richardxia:
Let's merge this in along with the frontend changes to test this on staging.

app/models/resource.rb
richardxia: This line is causing the TestCafe test for adding a new resource to fail, but I can't figure out how this used to work before. The error message is:

```rb
 NoMethodError (undefined method `latitude' for nil:NilClass)
```

However, I don't see where `address_latitude` or `address_longitude` were defined, and therefore how this handled the case where there is no latitude or longitude. Do you know where they were defined?

PR #472, opened 2019-11-17
Allow access to rails tools from terminal

alexanderturinske:
- previously, one had to ssh into the docker container to
  track byebug or run the rails console
- create a psudoterminal to the api docker container and
  allow input for it
- update readme to explain how to run the rails console and
  byebug
- https://stackoverflow.com/questions/31669226/rails-byebug-did-not-stop-application/32690885#32690885
- https://medium.com/@ddomachine/docker-and-ruby-on-rails-448ab8255f97

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,20 @@ $ docker-compose run --rm api rake db:create db:schema:load linksf:import
 $ docker-compose run -e STAGING_DB_PASSWORD=<...> --rm api rake db:setup db:import_staging
 ```
 
+### Running tools from the terminal
+Due to the DB and API being contained inside Docker containers, running ruby/rails specific tools without having to ssh into the containers involves specific steps
+#### Running the rails console from the terminal
+```sh
+# Start both the DB and API
+$ cd PATH/TO/askdarcel-api
+$ docker-compose run --rm api rails console
+```
+
+#### Running byebug from the terminal
+```sh
+# Start both the DB and API
+$ docker attach $(docker ps -aqf "name=askdarcel-api_api")
+```
 
 ## macOS-based Development Environment Not Using Docker
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -29,6 +29,10 @@ services:
     command: rails server --port=3000 --binding=0.0.0.0
     ports:
       - "3000:3000"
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -80,6 +80,20 @@ $ docker-compose run --rm api rake db:create db:schema:load linksf:import
 $ docker-compose run -e STAGING_DB_PASSWORD=<...> --rm api rake db:setup db:import_staging
 ```
 
+### Running tools from the terminal
+Due to the DB and API being contained inside Docker containers, running ruby/rails specific tools without having to ssh into the containers involves specific steps
+#### Running the rails console from the terminal
+```sh
+# Start both the DB and API
+$ cd PATH/TO/askdarcel-api
+$ docker-compose run --rm api rails console
```
richardxia: While I think it's useful to have this here, I'm not sure whether it belongs under a section like this.

The general rule of thumb is for any command that you would normally native run, just prefix it with `docker-compose run --rm api`.

Also, I wasn't exactly sure what you meant by "ssh into the containers". We don't literally SSH into the containers with `ssh`, but the closest thing to that is executing a command within the container while attaching your terminal's stdin/stdout. But that is exactly what you are doing here with this command, which seems to conflict with the sentence you wrote at the beginning of the section.

README.md
```diff
@@ -80,6 +80,20 @@ $ docker-compose run --rm api rake db:create db:schema:load linksf:import
...
+```
+
+#### Running byebug from the terminal
+```sh
+# Start both the DB and API
+$ docker attach $(docker ps -aqf "name=askdarcel-api_api")
```
richardxia: It might be worth describing this in a lot more detail. For people who aren't familiar with byebug, this doesn't actually teach anyone how to start byebug. What this command really does is attach your TTY and stdin to an already running Docker container which is assumed to be running the Rails web server. The comment above this is actually incorrect; this command will fail if the API server is not already running.

README.md
alexanderturinske: Good points. I have updated the wording to be, hopefully, correct and removed the contradictions.

I renamed the section to be `Developer Tools`; do you think that is a better section have.

README.md
alexanderturinske: I have added more information; let me know if you want me to add more.

PR #493, opened 2020-01-26
484 Add method to check for JWT tokens passed into the API

Raawr:


lgarofalo:
Looks good for this step, going to merge so we can validate in staging.

I'll follow up with signature verification.

PR #495, opened 2020-01-29
feat(auth): verify JWT signatures and update more admin endpoints to …

lgarofalo:
…do JWT check

I also added a ruby file formatting gem, and touched a couple files.

There are a few TODO's here, I will follow up with:

* Update dev workflow so that it doesn't 401 in dev without valid JWT for admin API
* Pull the JWT issuer public cert dynamically, probably on startup once.
* Format the whole repo using `rubo`
* Add full coverage of all admin operations
* Additional code clean-up / refactoring / learning ruby.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,7 +35,6 @@ gem 'jwt'
 gem 'lograge'
 
 gem 'prometheus-client'
-gem 'rufo'
 
 gem 'rspec-rails'
 
--- a/Gemfile
+++ b/Gemfile
@@ -35,6 +35,7 @@ gem 'jwt'
 gem 'lograge'
 
 gem 'prometheus-client'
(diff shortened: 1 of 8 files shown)


lgarofalo:
I'm finding that rufo and rubocop don't get along, I'll look for something else to format.  We should really get some auto formatting build into the repo.

richardxia:
Rubocop is honestly kind of a pain in the ass, and I've been a huge fan of code formatters in other languages, so I'd be down with replacing rubocop with rufo.

Alternatively, we could use both, but just disable anything in rubocop that conflicts with rufo, since rubocop also checks for things besides whitespace and parentheses.

lgarofalo:
I'm going to merge this for now, and follow with up improvements.

PR #497, opened 2020-02-05
update travis to only push tagged or latest images

lgarofalo:


richardxia:
@lgarofalo, I think this PR had an issue that wasn't obvious until I merged in https://github.com/ShelterTechSF/askdarcel-api/pull/506, but which isn't even visible until a PR is submitted directly to this repo rather than a fork (like https://github.com/ShelterTechSF/askdarcel-api/pull/507).

Despite the commit message you wrote here, this change only makes it so that the `TAG` shell variable is set to a non-empty value on tagged or `master` builds. However, Travis apparently still tries to build the Docker image and just fails. However, until my PR merged, the failures were silent.

Here's an example of a Travis CI build from _before_ my PR merged where both the `docker build` and the `docker push` commands failed, but Travis still passed: https://travis-ci.org/github/ShelterTechSF/askdarcel-api/builds/654765065#L1697

```
$ docker build -f Dockerfile -t $REPO:$TAG .
invalid argument "sheltertechsf/askdarcel-api:" for "-t, --tag" flag: invalid reference format
See 'docker build --help'.

$ docker push $REPO
The push refers to repository [docker.io/sheltertechsf/askdarcel-api]
An image does not exist locally with the tag: sheltertechsf/askdarcel-api

Done. Your build exited with 0.
```

My PR apparently escalates it to an actual failure: https://travis-ci.org/github/ShelterTechSF/askdarcel-api/builds/662647996?utm_source=github_status&utm_medium=notification

Let me attempt to fix this to combine both of our PRs properly.

PR #500, opened 2020-02-24
Set Environment Variable POSTGRES_HOST_AUTH_METHOD to trust

davidagustin:
For docker-compose.yml, postgres requires a password to be
set for docker to work. This change sets postgres to not
require a password.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,6 +2,8 @@ version: "3.5"
 
 services:
   db:
+    environment:
+      POSTGRES_HOST_AUTH_METHOD: trust
     image: "postgres:9.5"
     networks:
       - askdarcel


davidagustin:
The cache makes previous docker builds to work but a fresh build breaks it because of this change. This issue explains the problem: https://github.com/docker-library/postgres/issues/681

richardxia:
@lgarofalo, it looks like this PR is failing because it was submitted from a fork (@davidagustin has not yet been added to the ShelterTechSF Organization, and can't push branches directly to this repo). Specifically, it's hanging when it tries to run `docker login`, since the environment variables containing the Docker Hub secrets are not getting set, which is intentional on the Travis CI side for [security reasons](https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions). It therefore gets stock on the password prompt.

Two things:
1. Can we change the Travis CI builds to not attempt to log in and push Docker images to Docker Hub for PR builds and
2. Can we put double quotes around the `$DOCKER_USERNAME` and `$DOCKER_PASSWORD` variables so that if they evaluate to the empty string, the actually still pass the values of empty strings to the underlying commands? Normal shell variable expansion will just result in not passing in an argument at all, since variables are expanded and _then_ whitespace parsing happens. See https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_03_04.html for the gory details. If we quote the variables, then I believe this would at least force the command to fail faster rather than assume that we want to drop into an interactive prompt.

richardxia:
I fixed the things I described in #506, but I am unable to push to this branch because it doesn't look like you've granted access to it. I'm going to close this PR and just open a new one where your commit has been rebased on top of master.

PR #506, opened 2020-03-15
Only push Docker image on non-PR builds.

richardxia:
This is not only unnecessary to do on PR builds, but it also prevents builds on PRs coming from forks of the repository from succeeding, since the Docker Hub API keys are not available on forks.

This should fix the issue in https://github.com/ShelterTechSF/askdarcel-api/pull/500#issuecomment-590192394.

I am submitting this from a personal fork of askdarcel-api to test if that this does in fact work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,11 @@ script:
   - ./travis/postman.sh
 
 after_success:
-  - docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
   - export TAG=`if [[ -n "$TRAVIS_TAG" ]]; then echo "$TRAVIS_TAG"; else if [ "$SANITIZED_BRANCH" == "master" ]; then echo "latest"; fi; fi`
-  - docker build -f Dockerfile -t $REPO:$TAG .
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -40,7 +40,11 @@ script:
   - ./travis/postman.sh
 
 after_success:
-  - docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
   - export TAG=`if [[ -n "$TRAVIS_TAG" ]]; then echo "$TRAVIS_TAG"; else if [ "$SANITIZED_BRANCH" == "master" ]; then echo "latest"; fi; fi`
-  - docker build -f Dockerfile -t $REPO:$TAG .
-  - docker push $REPO
+  - docker build -f Dockerfile -t "$REPO:$TAG" .
+  - |
+    set -euo pipefail
```
richardxia: Apparently `set -u` does not work in Travis because one of _their_ built-in functions has an unbound variable. https://github.com/travis-ci/travis-ci/issues/5434

Will just remove this stuff...

PR #514, opened 2020-04-20
added lines of code to support deleting an existing address

azhelle16:
code to be reviewed
comments are to be removed once code is ready for prod

Name    :   Maricel Louise Sumulong
Date    :   April 19, 2020 PDT
E-mail  :   mllsumulong@gmail.com

### DIFF
--- a/app/controllers/change_requests_controller.rb
+++ b/app/controllers/change_requests_controller.rb
@@ -180,17 +180,34 @@ class ChangeRequestsController < ApplicationController
       phone.update field_change_hash
     elsif change_request.is_a? AddressChangeRequest
       puts "AddressChangeRequest"
+      # action = change_request.field_changes["action"]
+      action = "" # to be removed for end-to-end testing
+      #puts "ACTION: #{action}"
+      # change_request.field_changes.each do |fc|
+      #   puts "#{fc.field_name} ==== #{fc.field_value}"
+      # end
       address = Address.find(change_request.object_id)
 
-      begin
-        a = Geokit::Geocoders::GoogleGeocoder.geocode address.address_1 + "," + address.city + "," + address.state_province
-        field_change_hash["latitude"] = a.latitude
-        field_change_hash["longitude"] = a.longitude
-      rescue => error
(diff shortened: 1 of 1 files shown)


app/controllers/change_requests_controller.rb
```diff
@@ -180,17 +180,34 @@ def persist_change(change_request)
...
-      rescue => error
-        puts "google geocoding failed for address " + address.id.to_s + ": " + error.message
-      end
+      if action.downcase == "remove"
+        #address.destroy
+        #puts "ID: #{change_request.object_id.to_s}"
+        begin
+          address.destroy
+        rescue => error
+          puts error.message
```
richardxia, line 197: Just wondering, were you encountering errors trying to call `#destroy` on the Address? If so, what was the error?

app/controllers/change_requests_controller.rb
```diff
@@ -180,17 +180,34 @@ def persist_change(change_request)
       phone.update field_change_hash
     elsif change_request.is_a? AddressChangeRequest
       puts "AddressChangeRequest"
+      # action = change_request.field_changes["action"]
+      action = "" # to be removed for end-to-end testing
```
richardxia, line 184: The `action` is an enum on the [`ChangeRequest`](https://github.com/ShelterTechSF/askdarcel-api/blob/3ad4773352ddd5d0c089008197a1035d1279733c/app/models/change_request.rb#L5) model, meaning that once it's saved to the database, it is represented as an integer. This means that you should be able to check the enum value directly on the model instance, and not the `field_change_hash`, using either the name of the enum property or the special boolean methods for each enum value like so:

```rb
change_request.status  # Gives you a string holding "insert", "edit", or "remove"
change_request.remove? # Returns true if and only if the status is set to "remove"
```

See https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Enum.html for some docs on how Rails enums work.

richardxia:
Thanks for submitting this, this is a good start!

katerina-kossler:
Stale Branch - working on a new branch for 1319 so closing this and another out

PR #520, opened 2020-05-03
1343 create addresses endpoint

jjfreund:


app/controllers/change_requests_controller.rb
```diff
@@ -13,16 +13,28 @@ def create
   end
 
   def create_change_request
+
+    if params[:change_request][:action] == "0"
+      handle_insert
+    elsif params[:change_request][:action] == "1"
```
richardxia, line 19: Looks like we're going with the numbers (as strings) for the API. Should @azhelle16 also adopt this convention instead of the string "remove" in https://github.com/ShelterTechSF/askdarcel-api/pull/514 ?

app/controllers/change_requests_controller.rb
```diff
@@ -53,6 +65,68 @@ def create_change_request
     render status: :created, json: ChangeRequestsPresenter.present(@change_request)
   end
 
+  def handle_insert
+
+    change_request = params[:change_request]
+    if change_request[:type] == "addresses"
+      address = Address.new
+      addressattention = change_request[:field_changes][:attention]
```
richardxia, line 73: Typo? I think you're missing a `.` between "address" and "attention".

app/controllers/change_requests_controller.rb
```diff
@@ -53,6 +65,68 @@ def create_change_request
...
+      address.address_4 = change_request[:field_changes][:address_4]
+      address.city = change_request[:field_changes][:city]
+      address.state_province = change_request[:field_changes][:state_province]
+      address.postal_code = change_request[:field_changes][:postal_code]
+      address.country = change_request[:field_changes][:country]
+      address.online = change_request[:field_changes][:online]
+      address.region = change_request[:field_changes][:region]
+      address.name = change_request[:field_changes][:name]
+      address.description = change_request[:field_changes][:description]
+      address.transportation = change_request[:field_changes][:transportation]
```
richardxia, line 87: I think most of these could be done in a single, bulk statement using `assign_attributes`: https://stackoverflow.com/a/9751488

app/controllers/change_requests_controller.rb
```diff
@@ -53,6 +65,68 @@ def create_change_request
...
+      address.resource_id = change_request[:resource_id]
+
+      begin
+        a = Geokit::Geocoders::GoogleGeocoder.geocode address.address_1 + "," + address.city + "," + address.state_province
+        address.latitude = a.latitude
+        address.longitude = a.longitude
+      rescue => error
+        puts "google geocoding failed for address " + address.id.to_s + ": " + error.message
+      end
```
richardxia, line 96: This seems like something that would be good to refactor into a single method.

app/controllers/change_requests_controller.rb
```diff
@@ -13,16 +13,28 @@ def create
   end
 
   def create_change_request
+
+    if params[:change_request][:action] == "0"
+      handle_insert
+    elsif params[:change_request][:action] == "1"
+      handle_update 
+    else
+      puts 'unsupported action'
```
richardxia, line 22: Maybe this should be a 400 error?

Also, maybe this whole thing could be a `case` statement, and then @azhelle16 could slot her changes in pretty easily by adding a `when` clause for `"2"` (remove).

app/controllers/change_requests_controller.rb
```diff
@@ -13,16 +13,28 @@ def create
...
+
     if params[:resource_id]
       @change_request = ResourceChangeRequest.create(object_id: params[:resource_id], resource_id: params[:resource_id])
     elsif params[:service_id]
       @change_request = ServiceChangeRequest.create(object_id: params[:service_id], resource_id: Service.find(params[:service_id]).resource_id)
     elsif params[:address_id]
       @change_request = AddressChangeRequest.create(object_id: params[:address_id], resource_id: Address.find(params[:address_id]).resource_id)
     elsif params[:phone_id] || params[:type] == "phones"
-      resource_id = params[:parent_resource_id] || Phone.find(params[:phone_id]).resource_id
+      resource_id = params[:change_request][:parent_resource_id] || Phone.find(params[:phone_id]).resource_id
```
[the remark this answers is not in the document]
richardxia: Also, did you want this to be `resource_id` or `parent_resource_id`?

app/controllers/change_requests_controller.rb
richardxia: Actually, I take that back; can you make this default to `handle_insert` if the `action` isn't present. This is otherwise going to silently break every single frontend API call to the change requests since they aren't already passing in the `action`.

I'm also open to adding the `action` param to every call, but that's a lot of them to edit on the frontend.

jjfreund: yeah I can do that, but if you look at the postman tests a lot of the update client calls will have to be changed anyway

richardxia:
Rebased this on top of master in order to pick up https://github.com/ShelterTechSF/askdarcel-api/pull/519, so that I can see what the status of the Postman tests are, since I think I'm seeing some failures that I now believe are not due to local changes I made to this PR.

postman/AskDarcel%20API.postman_collection.json
```diff
@@ -1016,7 +1021,7 @@
 				],
 				"body": {
 					"mode": "raw",
-					"raw": "{  \n\t\"change_request\" : {\n\t\t\"city\" : \"San Francisco\",\n\t\t\"state_province\" : \"CA\",\n\t\t\"address_1\" : \"140 Market Street\"\n\t}\n}"
+					"raw": "{  \n\t\"change_request\" : {\n\t\t\"type\" : \"addresses\",\n\t\t\"action\" : \"0\", \n\t\t\"field_changes\" : {\n\t\t  \"city\" : \"San Francisco\",\n\t\t  \"state_province\" : \"CA\",\n\t\t  \"address_1\" : \"140 Market Street\"\n\t\t}\n\t}\n}"
```
richardxia, line 1024: This should have been `"1"`, not `"0"`, since it edits an existing address, not create a new address.

```suggestion
					"raw": "{  \n\t\"change_request\" : {\n\t\t\"type\" : \"addresses\",\n\t\t\"action\" : \"1\", \n\t\t\"field_changes\" : {\n\t\t  \"city\" : \"San Francisco\",\n\t\t  \"state_province\" : \"CA\",\n\t\t  \"address_1\" : \"140 Market Street\"\n\t\t}\n\t}\n}"
```

app/controllers/change_requests_controller.rb
richardxia: Changed this to be a string in https://github.com/ShelterTechSF/askdarcel-api/pull/520/commits/77e473c7706081d78a9261b966de991254158854.

app/controllers/change_requests_controller.rb
```diff
@@ -53,6 +71,74 @@ def create_change_request
...
+        address.longitude = a.longitude
+      rescue => error
+        puts "google geocoding failed for address " + address.id.to_s + ": " + error.message
+      end
+
+      address.save!
+
+      @change_request = AddressChangeRequest.create(object_id: address.id, resource_id: Address.find(address.id).resource_id)
+      @change_request.field_changes = field_changes
+      persist_change (@change_request)
```
richardxia, line 113: I just realized that this line is here, which is a bit confusing because we're saving the Address twice: once just a few lines higher up, and once here, after we construct the `@change_request` object.

@jjfreund, is there a reason you created the Address object above with all of the field changes saved before you call `persist_change`, which modifies the Address object again with the exact same field changes before saving it again?

It also looks like you're not saving the ChangeRequest's field changes to the database, which means that after it calls `persist_change`, it throws out the field change data from the ChangeRequest object after updating the object itself.

For consistence with the update methods and to eventually support the case where we gate edits by a moderator, I'd actually prefer to only create and save the address info after creating the ChangeRequest, using only info from the ChangeRequest. That would allow us to decouple in time the creation of the ChangeRequest and the approval of the ChangeRequest.

katerina-kossler:
Handled by another branch for 1343

PR #524, opened 2020-05-18
1498 new service and org discovery

jjfreund:


app/controllers/categories_controller.rb
```diff
@@ -1,6 +1,16 @@
 # frozen_string_literal: true
 
 class CategoriesController < ApplicationController
+  def show
+    category = Category.find(params[:id])
+    render json: CategoryPresenter.present(category)
+  end
+
+  def subcategories
+    categories = Category.where("id in (select child_id from category_relationships where parent_id=?)", params[:id])
```
richardxia, line 10: I think this just searches one sublevel. I feel that I saw in the past that we had like a 3 or 4-level deep list of subcategories, but is that not the case today or not necessary yet?

app/controllers/eligibilities_controller.rb
```diff
@@ -5,6 +5,15 @@ class EligibilitiesController < ApplicationController
   #
   # Return all eligibilities sorted by name in ascending order.
   def index
+    if params[:category_id]
+      eligibilities = Eligibility.where(
+        "id in (select distinct eligibility_id from eligibilities_services where service_id in " \
+        "(select service_id from categories_services where category_id=?))", params[:category_id]
```
richardxia, line 11: Is there a reason you did this as a bunch of subqueries instead of a flatter set of joins? I think in Rails, you could even do:

```ruby
Eligibility.joins({ services: :categories })
  .where({ services: { categories: { id: params[:category_id] } } })
```

PR #532, opened 2020-07-26
1523 added data for new pathways

jjfreund:
Data is contained in a sql script to be executed directly against the prod db. IDs of new categories are hardcoded to match hardcoded values in links on home page.

### DIFF
--- a/db/sql/pathways.sql
+++ b/db/sql/pathways.sql
@@ -1,0 +1,757 @@
+Insert into categories (id, created_at, updated_at, name, top_level, vocabulary, featured) values 
+(1000001, '06-JUN-20', '06-JUN-20', 'Covid-food', 'f', null, 'f');
+
+insert into eligibilities_services (eligibility_id, service_id) 
+values (1054, 305);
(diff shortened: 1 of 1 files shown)


lgarofalo:
Looks ok, reminder to run within a transaction
```
BEGIN;
\i thefile
```
commit or rollback etc.

jjfreund:
> I guess this LGTM, but is there a reason to do this in SQL and not Rails?
> 
> Also, an [UPSERT](https://wiki.postgresql.org/wiki/UPSERT) might be better to use here since it makes running this SQL idempotent, meaning that it's safe to run on a DB that already has this data.

It's slightly quicker to write in SQL, and no actual rails functionality is needed. As far as UPSERTS go, I would prefer the failure if someone runs it a second time, as that shouldn't occur and the script should be run as part of a transaction.

jjfreund:
> Looks ok, reminder to run within a transaction
> 
> ```
> BEGIN;
> \i thefile
> ```
> 
> commit or rollback etc.

Yep.

PR #534, opened 2020-09-06
backend for feedback mechanism

Akhtam:


app/models/feedback.rb
```diff
@@ -0,0 +1,4 @@
+class Feedback < ApplicationRecord
```
jjfreund, line 1: Since your object replaces ratings and reviews you should just clean up those unused objects as well.

app/models/feedback.rb
```diff
@@ -0,0 +1,4 @@
...
+    validates :rating, presence: true
+    belongs_to :feedbackable, polymorphic: true
```
jjfreund, line 3: Is it better to make this polymorphic rather than just having two optional relationships to resources and services? I guess there's some advantage in that you won't orphan an object with a relationship to nothing but it might make sql queries more difficult. I'm fine with it either way though. Also as you mentioned to me, if you remove the old tables then you could at least use a more sensible name "reviewable" rather than "feedbackable."

config/routes.rb
```diff
@@ -27,11 +27,13 @@
     resources :ratings, only: :create
```
jjfreund, line 27: the old routes for ratings etc could be cleaned up

config/routes.rb
```diff
@@ -27,11 +27,13 @@
...
     resources :change_requests, only: :create
     resources :services, only: :create
+    resources :feedbacks, only: [:create, :index]
   end
   resources :services do
     resources :ratings, only: :create
     resources :change_requests, only: :create
     resources :notes, only: :create
+    resources :feedbacks, only: :create
```
jjfreund, line 36: wouldn't we want a finder for feedbacks per service as well?

db/schema.rb
```diff
@@ -10,8 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_11_19_150125) do
-
+ActiveRecord::Schema.define(version: 2020_09_04_021230) do
```
jjfreund, line 14: The old objects could be cleaned up here too.

spec/controllers/feedbacks_controller_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper'
+
+RSpec.describe FeedbacksController, type: :controller do
```
jjfreund, line 3: Haha thank you for including tests we don't do nearly enough of these. However one thing we *do* usually do is to add basic postman tests that validate each endpoint. Those live in the postman directory and are used by TravisCI so please include some for your new routes.

jjfreund (changes requested):
Looks good! See comments for changes requested.

config/routes.rb
Akhtam: definitely I missed to add it on services, thank you

spec/controllers/feedbacks_controller_spec.rb
Akhtam: it was generated by Rails I can't get any credit for these lol

PR #537, opened 2020-09-24
1622 add sites table

katerina-kossler:
- Creates site table
- Creates an intermediate table for many-to-many relationship between sites and resources
- eventually want an enum for site_code but couldn't find a work around using active record


katerina-kossler:
Close to change site_code to a string

PR #538, opened 2020-09-28
1622 add sites table

katerina-kossler:
Add Sites table with site_code (string)
Add join table between Sites and Resources for many-to-many relationship

### DIFF
--- a/db/migrate/20200928034548_create_sites.rb
+++ b/db/migrate/20200928034548_create_sites.rb
@@ -1,7 +1,7 @@
 class CreateSites < ActiveRecord::Migration[5.2]
   def change
     create_table :sites do |t|
-      t.string :site_code
+      t.column :site_code, :string
     end
   end
 end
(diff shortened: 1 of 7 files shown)


jjfreund (changes requested):
The migration looks good. I think we also want a many-many relationship between sites and categories, as they will also be site-specific.

We also will want to add the rails domain object for sites (and modify resources & categories) but that can be done in a different commit.

db/migrate/20200928034548_create_sites.rb
```diff
@@ -0,0 +1,7 @@
+class CreateSites < ActiveRecord::Migration[5.2]
+  def change
+    create_table :sites do |t|
+      t.column :site_code, :integer, default: 'sfsg'
```
jjfreund, line 4: This should be a string no?

jjfreund (changes requested):

db/migrate/20200928034548_create_sites.rb
katerina-kossler: Yes!

jjfreund (approved):
lgtm! Not sure if we need the default value for site code but that shouldn't be harmful.

PR #539, opened 2020-10-01
1604 add covid jobs and covid financial categories

meltingmelon:

PR #541, opened 2020-10-19
1602 add covid shelter category

meltingmelon:
Wrote covid-shelter script to add Covid-shelter and its subcategories. Linked associated services and ran all commands in staging.

I did not add the covid-shelter category or its subcategories into fixture_populator.rb to prevent merging conflicts with #539


jjfreund (changes requested):
Can you also add this category to the local setup scripts?

PR #542, opened 2020-10-21
Build Site model and add many_to_many relationships for Resource and Category models

katerina-kossler:
Build Site model and add many_to_many relationships for Resource and Category models


lib/tasks/create_site.rake
```diff
@@ -0,0 +1,37 @@
...
+  # create site_code of 'sfsg'
+  desc 'Creates a site for sfsg'
+  task add_sfsg_site: :environment do
+    Site.transaction do
+      Site.find_or_create_by(site_code: 'sfsg')
+    end
+  end
+  
+  # to each resource, add a connection using resource_site
+  desc 'Adds the sfsg site to existing resources'
```
jjfreund, line 15: We might need to refine the business logic a bit here, as I think (whoops) some of the sffamilies orgs have already been added. We can do that in another PR though.

jjfreund (changes requested):
Looks good for the most part! You may need to add a couple new lines (and correct other stylistic issues) as our delinter, Rubocop, is complaining. You can run the delinter locally with the command rubocop --auto-correct.

PR #546, opened 2020-11-18
api-docs

lgarofalo:


postman/README.md
```diff
@@ -0,0 +1,9 @@
+#### API Documentation
+
+There is a new collection `Sheltertech_Data.postman_collection.json` that serves as a functional API doc for a read of set only endpoints.
+This uses the `Development` and `Staging` environments.  
+We can expand on this documentation to include the Change Request and Admin interfaces.
+
+This new collection syncs with Postman's cloud hosting and reflects at https://documenter.getpostman.com/view/13231981/TVYF8eMY
+
+To get this to synchronize you must be logged into Postman locally and make edits to the file.
```
lgarofalo, line 9: I need to migrate this into a "Team" shared space, working on that today and will add details here.

varthraeis: Sounds like we don't have to do this necessarily (per discussion on Sunday), but it'd be nice if we could give others access like myself. Is that something that would require another PR? Otherwise I don't think there's a ton of harm in merging this to my untrained eye.

drcaramelsyrup: Sounds like we don't have to do this necessarily (per discussion on Sunday), but it'd be nice if we could give others access like myself. Is that something that would require another PR? Otherwise I don't think there's a ton of harm in merging this to my untrained eye.

lgarofalo:
Updating with short guide to access team workspace.

PR #549, opened 2020-12-11
Texting

fayceltouili:


app/controllers/textings_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+        if texting.save
+          render status: :ok, json: { message:'Message sent' }
+        else
+          render status: :bad_request
+        end
+
+      end
+
+    else
+      render status: :bad_request, json: { error: response['message'] }
```
richardxia, line 44: Two comments here:

1. I don't think it's necessarily a good idea to return Textellent's error message to our web app client request. This would expose the internals of the interaction between our web app API server and Textellent's API, which could possibly contain sensitive information that we wouldn't want everyone to be able to see. Anything that is returned to the SF Service Guide web client could be visible to anyone who happens to have the Developer Tools Console open.

    We could probably just log an error locally, such as with [Sentry](https://sentry.io/welcome/), and just return a generic error to the API client. We currently use Sentry to log our errors, and we're still using their older library named Raven: https://github.com/getsentry/sentry-ruby/tree/3f37864306fba6dc0413105cf4620a11b09cda2b/sentry-raven. I _think_ you can use `Raven.capture_message()`, although we apparently don't have an example of this on our site, and I'm finding it difficult to find documentation on Raven, since Sentry is pushing people to use their new client library.

2. Could you rearrange your code into the form of an [early return](https://www.itamarweiss.com/personal/2018/02/28/return-early-pattern.html)? You can see the linked article or other articles on the web for more details, but it's often difficult to keep track of heavily nested if statements. In cases like you have, where one branch of the `if/else` is much smaller than the other, you may actually want to handle that first and return early, which allows the rest of the function be the non-error case and not be indented.

    More specifically, I would ask that you restructure your `create` method to look more like the following:

    ```rb
    response = post_textellent(get_db_data)

    if response['status'] != 'success'
      render status: :bad_request, json: { error: response['message'] }
      return
    end

    # Rest of the non-error case just goes here
    recipient = TextingRecipient.find_by(phone_number: texting_params[:phone_number])
    ...
    ```

app/controllers/textings_controller.rb
```diff
@@ -0,0 +1,108 @@
+class TextingsController < ApplicationController
+
+  def create
+    # Make a request to Textellent API. If request successful we update our DB
+    response = post_textellent(get_db_data)
+
+    if response['status'] == 'success'
+      recipient = TextingRecipient.find_by(phone_number: texting_params[:phone_number])
```
richardxia, line 8: So even if we do find a TextingRecipient with a matching phone number in the database, if for some reason the person's name changes, we may want to update the model in our database. This could happen in the very unlikely case where someone changes their phone number, and someone new gets that number, and both the old and the new owners of the number try to use this, but it could also happen if someone changes their name (e.g. they get married).

I know that this isn't the most likely thing to happen, but if it's only just a bit extra work, I think it'd be good to do it to future-proof the app.

app/controllers/textings_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+    service = Service.find(texting_params[:service_id])
+    category_ids = service.category_ids
+    resource = Resource.find(service.resource_id)
+    address = resource.addresses
+    phone = resource.phones[0].number
+
+    categories = []
+    category_ids.each do |category_id|
+      categories << Category.find(category_id).name
+    end
```
richardxia, line 63: Nit: I prefer to use a more functional programming approach of doing this by using the [`#map`](https://ruby-doc.org/core-2.7.0/Array.html#method-i-map) method on arrays. This would look like:

```suggestion
    categories = category_ids.map do |category_id|
      Category.find(category_id).name
    end
```

Also, does `Service` have a `categories` property, rather than the `categories_id` property? You may be able to save some unnecessary database calls by prefetching all the categories and then accessing their names directly. I haven't tested this code, but I think this could look like the following:

```rb
service = Service.includes(:categories).find(texting_params[:service_id])

categories = service.categories.map { |c| c.name }
```

See https://engineering.gusto.com/a-visual-guide-to-using-includes-in-rails/ for a good guide on how `.includes()` and prefetching works in Rails.

app/controllers/textings_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    data = {
+    "firstName" => recipient_name,
```
richardxia, line 66: Nit: Can we indent all the keys by one level so that they are visually indented, rather than vertically lined up with all of the top-level statements?

app/controllers/textings_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+    "mobilePhone" => phone_number,
+    "phoneAlternate": "",
+    "phoneHome" => "",
+    "phoneWork" => "",
+    "tags" => categories,
+    "engagementType" => "Resource Info",
+    "engagementInfo" => {
+        "Org_Name" => service.name,
+        "Org_Address1" => address[0].address_1,
+        "Org_Address2" => address[0].address_2,
```
richardxia, line 77: One thing you may want to be careful about is that not all resources have the second address line filled out. I don't recall off the top of my head whether this means the values are `nil` or if they are the empty string (`""`) from the database, but you may want to double check that we are massaging this in whatever way necessary for Textellent's API.

app/controllers/textings_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  # Generate a data object to send to Textellent API
+  def get_db_data
+    recipient_name = texting_params[:recipient_name]
+    phone_number = texting_params[:phone_number]
+    service = Service.find(texting_params[:service_id])
+    category_ids = service.category_ids
+    resource = Resource.find(service.resource_id)
+    address = resource.addresses
+    phone = resource.phones[0].number
```
richardxia, line 58: You may want to be careful that the resource actually has a phone number, since not all resources will have a phone number attached.

app/controllers/textings_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+    "lastName" => "",
+    "mobilePhone" => phone_number,
+    "phoneAlternate": "",
+    "phoneHome" => "",
+    "phoneWork" => "",
+    "tags" => categories,
+    "engagementType" => "Resource Info",
+    "engagementInfo" => {
+        "Org_Name" => service.name,
+        "Org_Address1" => address[0].address_1,
```
richardxia, line 76: This is less common, but some resources do not have any addresses. This is usually because the resource has no physical location and are just some online resource, but sometimes this is also because the data is incomplete in the database. You may want to be careful that you don't trigger an exception when trying to index into a zero-length array.

spec/factories/texting_recipients.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :texting_recipient do
+    recipient_name { "MyString" }
+    phone_number { "MyString" }
```
richardxia, line 4: It might be good to generate fake data that looks more like an actual phone number. We use a library named [`Faker`](https://github.com/faker-ruby/faker) which is useful for generating different kinds of structured data, and they even have a method for generating fake phone numbers. See https://github.com/ShelterTechSF/askdarcel-api/blob/de32f656c19116507d0ab9f35a58eadd88583f12/spec/factories/phones.rb#L11 for where we currently use that function for other testing factories.

app/controllers/textings_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+        texting = Texting.new(
+          texting_recipient_id: recipient.id,
+          service_id: texting_params[:service_id]
+          )
+
+        if texting.save
+          render status: :ok, json: { message:'Message sent' }
+        else
+          render status: :bad_request
+        end
```
richardxia, line 39: This branch of the nested `if/else` statements is very similar to the other branch where you create the `Texting` object. I think you could actually unify them into a single branch if you just pull out the conditional for when the TextingRecipient doesn't yet exist.

Here's a rough sketch of the change I'm suggesting, with the details omitted:

```rb
# The current code you have:

recipient = try_to_find_recipient()
if recipient
  texting = create_new_texting()
else
  recipient = create_new_recipient()
  texting = create_new_texting()
end


# Change that I am suggesting:
recipient = try_to_find_recipient()
if not recipient:
  recipient = create_new_recipient()

texting = create_new_texting()
```

richardxia (commented):
Thanks for splitting out the data model into two separate tables! This looks in line with what I was suggesting last time.

This looks pretty good to me, and I don't see any structural issues with what you have, so most of my comments are mostly localized improvements that we could make.

db/migrate/20201211012159_create_texting_recipients.rb
```diff
@@ -0,0 +1,10 @@
+class CreateTextingRecipients < ActiveRecord::Migration[5.2]
+  def change
+    create_table :texting_recipients do |t|
+      t.string :recipient_name
+      t.string :phone_number
```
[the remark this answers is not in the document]
fayceltouili: ` recipient_name` is optional, but you'r right we need to make `phone_number` non-nullable

app/controllers/textings_controller.rb
```diff
@@ -0,0 +1,122 @@
...
+    recipient_name = texting_params[:recipient_name]
+    phone_number = parse_phone_number
+    service_id = texting_params[:service_id]
+
+    # Make a request to Textellent API. If request successful we update our DB
+    data = get_db_data(recipient_name, phone_number, service_id)
+    response = post_textellent(data)
+
+    if response['status'] != 'success'
+      puts response['message']
```
richardxia, line 14: Nit: Let's not leave in print statements in the code. This ends up clogging the console if everyone just put their prints in the code. If you want to log this, you should use the Rails logger: https://guides.rubyonrails.org/debugging_rails_applications.html#the-logger

app/controllers/textings_controller.rb
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    data
+  end
+
+  # handling the post request to Textellent API.
+  def post_textellent(data)
+    header = {
+      'Content-Type' => 'application/json',
+      'authCode' => Rails.configuration.x.textellent.api_key
```
richardxia, line 110: You might want to add a check at the top of this function to make sure that the api_key is actually present, and if not, just dummy out a response. There's no need to make an API request to their actual API if an API key is not even sent, which is going to be the case for most developers on this project as well as the CI system.

richardxia (approved):
Thanks for making the changes, this looks much cleaner! I have a few small comments, but their non-blocking, and if necessary, we can fix them after deploying this to staging.

lgarofalo:
I added a commit to fix the SSL certificate issue we ran into, the endpoint for textellent returns a bad chain with expired intermediate and root, and the httpclient in ruby was using a truststore with the expired root.

This leaves one rspec test failing, I think that rspec test should be set up to mock the external endpoint though.

PR #550, opened 2020-12-14
1625 modify resource creation to optionally include one or more sites

katerina-kossler:
Added 'sfsg' and 'sffamilies' sites.

Added default connection of 'sfsg' site with resource creation and the ability to attach other sites to a resource (so far only needs support for 'sffamilies' through site_id of 2 or 'sfsg'/1 otherwise)

Support for postman tests has been added for adding resources with and without sites as well.


app/controllers/resources_controller.rb
```diff
@@ -79,6 +79,11 @@ def fix_resources(resources)
       r.addresses.each do |a|
         fix_lat_and_long(a)
       end
+
+      if r.sites.length.zero?
```
richardxia, line 83: Sorry for the late, drive-by commenting here, but the idiomatic way to check for an empty array in Ruby is to call the [`.empty?`](https://ruby-doc.org/core-2.7.0/Array.html#method-i-empty-3F) method.

PR #553, opened 2021-01-04
1808 modify search endpoint

jjfreund:


app/controllers/services_controller.rb
```diff
@@ -11,14 +11,66 @@ class ServicesController < ApplicationController
...
+  # rubocop:disable MethodLength
+  # rubocop:disable PerceivedComplexity
+  def find_by_categories(categories_id_string, eligibilities_id_string)
+    query_string = ""
+    if categories_id_string.include? ","
+      category_ids = categories_id_string.split ","
+      first_cat = true
+      category_ids.each do |cat_id|
+        if first_cat
+          query_string = "services.id in (select service_id from categories_services where category_id=" + cat_id + ")"
```
richardxia, line 32: This will expose us to a SQL injection attack. Can this be written using the ActiveRecord methods that properly escape external values? Or if you have to drop into raw SQL, at least use the sanitization methods available in ActiveRecord: https://api.rubyonrails.org/v5.1.0/

app/controllers/services_controller.rb
```diff
@@ -11,14 +11,66 @@ class ServicesController < ApplicationController
...
+          first_cat = false
+        else
+          query_string = query_string +
+                         " and services.id in (select service_id from categories_services where category_id=" + cat_id + ")"
+        end
```
richardxia, line 37: I think what you're trying to do might be simpler using an approach like this: https://stackoverflow.com/a/15977167

I think that'd avoid having to dynamically generate a query with an arbitrary number of expressions, since you can push everything into that inner join with a `HAVING` clause. Man, I haven't used the `HAVING` keyword in forever...

app/controllers/services_controller.rb
jjfreund: Ooh that's a good point. I could just check if all of the incoming IDs are integers for example.

app/controllers/services_controller.rb
jjfreund: Not sure how that would make things easier. I'll contact you to get more information.

PR #558, opened 2021-01-17
1635: Backfill missing schedules to fix Edit page hanging bug

quanhuynh:
Clubhouse URL: https://app.clubhouse.io/sheltertech/story/1635/editing-certain-organizations-hangs

Added a rake task to backfill missing schedules for services without one.

Before query (194 services missing schedules)
Screen Shot 2021-01-17 at 3 26 32 PM

After query
Screen Shot 2021-01-17 at 3 26 40 PM

A problematic edit page now fixed on local ([same page in staging](https://staging.askdarcel.org/organizations/977/edit))
Screen Shot 2021-01-17 at 3 26 59 PM

### DIFF
--- a/lib/tasks/onetime.rake
+++ b/lib/tasks/onetime.rake
@@ -717,4 +717,24 @@ namespace :onetime do
       child_id: child_id
     )
   end
+
+  # Services should always have a corresponding schedule objects, even if they inherit from resource
+  # This task backfills missing schedule objects for some services
+  desc 'Add schedule objects to services that are missing them'
+  task add_missing_schedule_objects: :environment do
+    Schedule.transaction do
+      # Get services with no schedule objects
+      rogue_services = Service.joins('LEFT JOIN schedules '\
+          'ON services.id = schedules.service_id '\
+          'WHERE schedules.id is NULL')
+      rogue_services.each do |service|
+        puts format(
+          'Creating schedule object for service %<service_id>i',
+          service_id: service.service_id
+        )
+        Schedule.create(resource_id: nil, service_id: service.id)
+      end
+      puts format('Created %<num>i missing schedules', num: rogue_services.length)
+    end
+  end
 end


lib/tasks/onetime.rake
```diff
@@ -717,4 +717,24 @@ namespace :onetime do
...
+
+  # Services should always have a corresponding schedule objects, even if they inherit from resource
+  # This task backfills missing schedule objects for some services
+  desc 'Add schedule objects to services that are missing them'
+  task add_missing_schedule_objects: :environment do
+    Schedule.transaction do
+      # Get services with no schedule objects
+      rogue_services = Service.joins('LEFT JOIN schedules '\
+          'ON services.id = schedules.service_id '\
+          'WHERE schedules.id is NULL')
```
richardxia, line 729: This is fine, especially since this is just a one-time task, but for future reference, you can do multiline strings in Ruby with the special `%` strings: https://stackoverflow.com/a/18478045, https://en.wikibooks.org/wiki/Ruby_Programming/Syntax/Literals#The_%_Notation

It'd look kind of like:

```rb
      rogue_services = Service.joins(%{
        LEFT JOIN schedules 
          ON services.id = schedules.service_id
          WHERE schedules.id is NULL
      %})
```

Note that this will insert actual newline characters and excessive spaces in the string due to the indentation, but it doesn't matter for SQL statements, since SQL doesn't care about having the extra whitespace. Also for future reference, if you did care about excessive spaces, you could also use the Ruby 2.3.0 syntax for squiggly heredocs: https://en.wikibooks.org/wiki/Ruby_Programming/Syntax/Literals#The_%_Notation

richardxia (commented):
This PR overall looks good to me, but could you or @jjfreund remind me how we represent services that should inherit their schedule from their associated resource/organization? I'm a bit hazy on the details, but I thought we interpreted null schedules on services to mean "inherit from resource". Could you double check that the frontend code still properly does the inheritance even when the schedule exists on the service?

quanhuynh:
> This PR overall looks good to me, but could you or @jjfreund remind me how we represent services that should inherit their schedule from their associated resource/organization? I'm a bit hazy on the details, but I thought we interpreted null schedules on services to mean "inherit from resource". Could you double check that the frontend code still properly does the inheritance even when the schedule exists on the service?

Yeah great question. You are thinking of `schedule_days` objects. Those are indeed null for services that inherit. The schedule objects only links to a resource/service (so it's more like metadata). The actual schedule data lives in schedule_days.
And as far as i can tell from spot checking the FE still does inheritance properly.

lib/tasks/onetime.rake
quanhuynh: oh great! thank you 😃

richardxia:
> Yeah great question. You are thinking of `schedule_days` objects. Those are indeed null for services that inherit. The schedule objects only links to a resource/service (so it's more like metadata). The actual schedule data lives in schedule_days.
> And as far as i can tell from spot checking the FE still does inheritance properly.

Cool, and yeah, I checked with Joe over Slack about inheritance working with empty schedules, so this all LGTM!

PR #559, opened 2021-02-01
categories are retrieved based on site_id

meltingmelon:
Updated index code to now show categories based on site_id.

### DIFF
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -12,11 +12,17 @@ class CategoriesController < ApplicationController
   end
 
   def index
-    categories = Category.order(:name)
+    # categories = Category.order(:name)
     # Cast:
     #   nil and '' -> nil
     #   '0', 'false', 'False', 'f', etc. -> false
     #   Almost everything else -> true
+    categories = if params[:site_id]
+                   Site.find(params[:site_id]).categories
+                 else
+                   Site.find(0).categories #sfsg
+                 end
+
     top_level = ActiveRecord::Type::Boolean.new.cast(params[:top_level])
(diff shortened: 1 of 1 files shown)


app/controllers/categories_controller.rb
```diff
@@ -12,11 +12,17 @@ def subcategories
   end
 
   def index
-    categories = Category.order(:name)
+    # categories = Category.order(:name)
```
drcaramelsyrup, line 15: Please remove if no longer necessary.

app/controllers/categories_controller.rb
```diff
@@ -12,11 +12,17 @@ def subcategories
...
     # Cast:
     #   nil and '' -> nil
     #   '0', 'false', 'False', 'f', etc. -> false
     #   Almost everything else -> true
+    categories = if params[:site_id]
+                   Site.find(params[:site_id]).categories
+                 else
+                   Site.find(0).categories #sfsg
```
drcaramelsyrup, line 23: `sfsg` is not necessarily site_id 0 - a safer way would be using `find_by(site_code: 'sfsg')`.

app/controllers/categories_controller.rb
```diff
@@ -12,11 +12,17 @@ def subcategories
...
+                 end
+
     top_level = ActiveRecord::Type::Boolean.new.cast(params[:top_level])
     categories = categories.where(top_level: top_level) unless top_level.nil?
```
drcaramelsyrup, line 27: Can we preserve the previous commented-out behavior by ordering by name after filtering by site?

drcaramelsyrup (changes requested):

PR #561, opened 2021-02-05
1837-Modify_db_nonCovid_categories_ontop

iopkelvin:
modified fixture_populator.rb (local) and create sql to delete covid subcategories (prod)


lgarofalo:
We will need to run this sql file when we deploy, correct?

iopkelvin:
yes. Does the rb file also need to be run? 
Local hasn't been updated.

PR #566, opened 2021-02-10
1836 - Add new API endpoint to get top level categories and their children

quanhuynh:
this work adds a new categories endpoint to get the top level categories and their immediate children
the response looks like this
Screen Shot 2021-02-09 at 5 31 25 PM


lgarofalo:
Is there a customer that plans to use this, just wondering where it will be used, we can include it in API doc work as well.

quanhuynh:
Yeah it's part of the work for admin interface. The ask from product is to have a 'portal' to be able to edit/add categories. This is the endpoint to grab them to display in frontend. This is only for internal Sheltertech use only though.

lgarofalo (approved):

PR #569, opened 2021-02-15
Order /services search results by number of matching tags

drcaramelsyrup:
This is not strictly required for SFFamilies launch, but it is a highly requested feature from their team.

Right now the services do not have any ordering when returned. There's no strict ordering preference the SFFamilies team has agreed on other than that they should be ranked by the number of matching tags (categories / eligibilities) - the search endpoint returns services that match _any_ of the tags currently, so this is just putting the ones that match _n_ categories ahead of the ones that match _n-1_ categories, etc.

This only increases the overall duration slightly by around a half second (the db query is not the most expensive part of the request handler by far).

```
http://localhost:3000/services?category_id=1,2,3
current master state, with resource lite:
api_1      | method=GET path=/services format=html controller=ServicesController action=index status=200 duration=13544.64 view=454.95 db=92.69
api_1      | method=GET path=/services format=html controller=ServicesController action=index status=200 duration=11526.69 view=439.81 db=51.39
api_1      | method=GET path=/services format=html controller=ServicesController action=index status=200 duration=12184.17 view=439.41 db=51.45

ordered categories:
api_1      | method=GET path=/services format=html controller=ServicesController action=index status=200 duration=13774.36 view=629.90 db=63.84
api_1      | method=GET path=/services format=html controller=ServicesController action=index status=200 duration=12510.76 view=636.01 db=60.30
api_1      | method=GET path=/services format=html controller=ServicesController action=index status=200 duration=12573.40 view=631.08 db=63.85
```

### DIFF
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -30,24 +30,13 @@ class ServicesController < ApplicationController
   # Include services if they:
   # - are part of a resource matching the requested site
   # - have any of the requested tags
+  # Sort by relevancy (i.e. number of matching tags).
   def find_by_category(category_id_string, site_id)
-    services.includes(
-      resource: [
-        :addresses, :phones, :categories, :notes,
-        schedule: :schedule_days,
-        ratings: [:review]
-      ]
-    ).where(by_category_join_string, (category_id_string.split ","), site_id)
+    find_by_tag("categories_services", "category_id", category_id_string, site_id)
   end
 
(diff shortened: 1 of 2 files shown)


drcaramelsyrup:
Updated with alphabetical sorting as a secondary sort order. Also added a fix to staging imports if the tables are new.

jjfreund (approved):
lgtm -- failing test unrelated.

PR #571, opened 2021-03-09
1343 create change request new address

katerina-kossler:
Added support for adding a New Address to a Resource through the change_requests endpoint.
Requested are expected to have the required fields to make an address in a 'change_request' hash along with the 'type' of change request and the 'parent_resource_id':
```
{  
  "change_request": {
        "address_1":"601 4th Street",
        "city":"San Francisco",
        "state_province":"CA",
        "country":"USA",
        "postal_code":"49032"
  },
  "type": "addresses",
  "parent_resource_id": 1
}
```
Refactored support code for adding a New Phone to a Resource.
Updated Postman tests to include support for testing 'Add New Address to Resource' ('Add New Phone to Resource' was already supported).
Exposed a route on change_requests to create.

### DIFF
--- a/app/controllers/change_requests_controller.rb
+++ b/app/controllers/change_requests_controller.rb
@@ -19,8 +19,8 @@ class ChangeRequestsController < ApplicationController
       @change_request = ServiceChangeRequest.create(object_id: params[:service_id], resource_id: Service.find(params[:service_id]).resource_id)
     elsif params[:address_id]
       @change_request = AddressChangeRequest.create(object_id: params[:address_id], resource_id: Address.find(params[:address_id]).resource_id)
-    elsif params[:phone_id] || params[:type] == "phones"
-      resource_id = params[:parent_resource_id] || Phone.find(params[:phone_id]).resource_id
+    elsif params[:phone_id]
+      resource_id = Phone.find(params[:phone_id]).resource_id
       @change_request = PhoneChangeRequest.create(object_id: params[:phone_id], resource_id: resource_id)
     elsif params[:schedule_day_id] || params[:type] == "schedule_days"
       schedule = nil
@@ -41,6 +41,8 @@ class ChangeRequestsController < ApplicationController
       else
         @change_request = NoteChangeRequest.create(object_id: params[:note_id], resource_id: Service.find(note.service_id).resource_id)
       end
+    elsif params[:parent_resource_id] && params[:type]
+      @change_request = handle_insert_change_request
     else
       render status: :bad_request
       return
@@ -51,6 +53,83 @@ class ChangeRequestsController < ApplicationController
     persist_change (@change_request)
 
     render status: :created, json: ChangeRequestsPresenter.present(@change_request)
+  end
+
+  def geocode_address(address)
+    begin
+      return Geokit::Geocoders::GoogleGeocoder.geocode address.address_1 + "," + address.city + "," + address.state_province
+    rescue => error
+      puts "google geocoding failed for address " + address.id.to_s + ": " + error.message
+    end
+  end
+
+  def handle_insert_change_request
+    resource_id = params[:parent_resource_id]
+    change_request = params[:change_request]
+    case params[:type]
+    when "addresses"
+      puts "AddressInsertChangeRequest"
+      address = Address.new
+      permitted_fields = [
+        :attention,
+        :address_1,
+        :address_2,
+        :address_3,
+        :address_4,
+        :city,
+        :state_province,
+        :postal_code,
+        :country,
+        :online,
+        :region,
+        :name,
+        :description,
+        :transportation
+      ]
+      required_fields = [
+        :address_1, 
+        :city, 
+        :state_province, 
+        :postal_code, 
+        :country
+      ]
+      change_request_params = change_request.require(:field_changes).permit(permitted_fields)
+      change_request_params.require(required_fields)
+
+      address.attributes = change_request_params
+      address.resource_id = resource_id
+      
+      a = geocode_address address
+      unless a.nil?
+        address.latitude = a.latitude
+        address.longitude = a.longitude
+      end
+
+      address.save!
+      AddressChangeRequest.create(object_id: address.id, resource_id: Address.find(address.id).resource_id)
+      
+    when "phones"
+      puts "PhoneInsertChangeRequest"
+      phone = Phone.new
+      permitted_fields = [
+        :description,
+        :number,
+        :service_type,
+        :service_id,
+        :contact_id,
+        :language_id
+      ]
+      required_fields = [:number, :service_type] 
+      change_request_params = change_request.require(:field_changes).permit(permitted_fields)
+      change_request_params.require(required_fields)
+      phone.attributes = change_request_params
+      phone.resource_id = resource_id
+
+      phone.save!
+      PhoneChangeRequest.create(object_id: phone.id, resource_id: Phone.find(phone.id).resource_id)
+    else
+      render status: :precondition_failed
+    end
   end
 
   def index
@@ -169,11 +248,7 @@ class ChangeRequestsController < ApplicationController
       note.update field_change_hash
     elsif change_request.is_a? PhoneChangeRequest
       puts "PhoneChangeRequest"
-      if change_request.object_id
-        phone = Phone.find(change_request.object_id)
-      else
-        phone = Phone.new(resource_id: change_request.resource_id, service_type: "")
-      end
+      phone = Phone.find(change_request.object_id)
       if field_change_hash["number"]
         field_change_hash["number"] = Phonelib.parse(field_change_hash["number"], "US").full_e164
       end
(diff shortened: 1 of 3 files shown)


app/controllers/change_requests_controller.rb
```diff
@@ -13,15 +13,26 @@ def create
   end
 
   def create_change_request
+    puts "#{params}"
```
jjfreund, line 16: Please remove all puts that are not necessary for the ongoing operation of the system. Statements like this without context will clutter logs and do not provide long term use.

app/controllers/change_requests_controller.rb
```diff
@@ -13,15 +13,26 @@ def create
...
+    puts "type: #{params[:type]}"
+    puts "parent: #{params[:parent_resource_id]}"
     if params[:resource_id]
       @change_request = ResourceChangeRequest.create(object_id: params[:resource_id], resource_id: params[:resource_id])
     elsif params[:service_id]
       @change_request = ServiceChangeRequest.create(object_id: params[:service_id], resource_id: Service.find(params[:service_id]).resource_id)
-    elsif params[:address_id]
-      @change_request = AddressChangeRequest.create(object_id: params[:address_id], resource_id: Address.find(params[:address_id]).resource_id)
+    elsif params[:address_id] || params[:type] == "addresses"
+      if params[:parent_resource_id]
```
jjfreund, line 24: Shouldn't you invoke the create if the params[:type] is set and params[:address_id] is absent? Or, better, take advantage of the action type?

jjfreund (changes requested):
Please remove the puts statements that don't yield useful long-term logging; I also had a question about the conditional logic that determines whether or not the handle_insert_change_request gets called.

app/controllers/change_requests_controller.rb
```diff
@@ -13,15 +13,26 @@ def create
...
+        @change_request = handle_insert_change_request
+      else
+        @change_request = AddressChangeRequest.create(object_id: params[:address_id], resource_id: Address.find(params[:address_id]).resource_id)
+      end
     elsif params[:phone_id] || params[:type] == "phones"
-      resource_id = params[:parent_resource_id] || Phone.find(params[:phone_id]).resource_id
-      @change_request = PhoneChangeRequest.create(object_id: params[:phone_id], resource_id: resource_id)
+      if params[:parent_resource_id]
```
jjfreund, line 30: Same as above ^^

app/controllers/change_requests_controller.rb
```diff
@@ -53,6 +64,82 @@ def create_change_request
     render status: :created, json: ChangeRequestsPresenter.present(@change_request)
   end
 
+  def  handle_insert_change_request
+    change_request = params[:change_request]
```
jjfreund, line 68: Also can you work with @richardxia whether or not you want to keep the existing structure that nests the properties to change as "field_changes;" your method currently does not do that.

richardxia: I could go either way. The older PR for this feature used [`field_changes`](https://github.com/ShelterTechSF/askdarcel-api/pull/520/files#diff-d111eb9645f57db8b0f6869b5975b9a1c22a7366882e4464378cdbba74d35b74R80) for some reason, but it looks like none of the existing endpoints actually expect data to be sent that way, so maybe we don't need to use `field_changes`.

It's pretty easy for me to change the frontend code to work either way, so just let me know what you decide.

I _think_ the reason the old PR did it that way is because the one thing that was in `:change_request` but not in `:field_changes` was the `:action` param, which I think is still useful to have.

app/controllers/change_requests_controller.rb
katerina-kossler: @richardxia I'm happy to start using the `:action` param! One question I had was if I should also require / look for the `:action` param for adding a phone to a resource. This currently works very similarly to addresses.

richardxia: I think it'd be a good idea to look for it when adding resources, but let's not require it for now, and just default to `:edit`. That way, we have the flexibility of updating the frontend site at a later date, instead of requiring it to be updated right now.

katerina-kossler: @richardxia I'm running into an issue where when I try to set the `:action` param I can't seem to override `create`: 
```
{  
    "action": "edit",
    "change_request": {
        "number": "(415) 555-5555",
        "service_type": "general inquiries"
    },
    "type": "phones",
    "parent_resource_id": 1
}
```
results in params of : 
```
{"action"=>"create", "change_request"=>{"number"=>"(415) 555-5555", "service_type"=>"general inquiries"}, "type"=>"phones", "parent_resource_id"=>1, "controller"=>"change_requests"}
```
Any ideas on what I'm missing here?

app/controllers/change_requests_controller.rb
katerina-kossler: Going ahead with changing the structure to use `fiend_changes` and `action` in the change request for both adding new addresses and phones to existing resources!

app/controllers/change_requests_controller.rb
```diff
@@ -19,9 +19,9 @@ def create_change_request
       @change_request = ServiceChangeRequest.create(object_id: params[:service_id], resource_id: Service.find(params[:service_id]).resource_id)
     elsif params[:address_id]
       @change_request = AddressChangeRequest.create(object_id: params[:address_id], resource_id: Address.find(params[:address_id]).resource_id)
-    elsif params[:phone_id] || params[:type] == "phones"
-      resource_id = params[:parent_resource_id] || Phone.find(params[:phone_id]).resource_id
-      @change_request = PhoneChangeRequest.create(object_id: params[:phone_id], resource_id: resource_id)
+    elsif params[:phone_id]
+        resource_id = Phone.find(params[:phone_id]).resource_id
+        @change_request = PhoneChangeRequest.create(object_id: params[:phone_id], resource_id: resource_id)
```
richardxia, line 24: Nit: These two lines should be dedented by one level.

```suggestion
      resource_id = Phone.find(params[:phone_id]).resource_id
      @change_request = PhoneChangeRequest.create(object_id: params[:phone_id], resource_id: resource_id)
```

app/controllers/change_requests_controller.rb
```diff
@@ -148,6 +223,7 @@ def reject
   private
 
   def persist_change(change_request)
+    # need to figure out what my object_id is
```
richardxia, line 226: For new objects (e.g. "insert" actions), you won't have an object ID yet. Creating the new object should automatically assign it an ID.

richardxia (commented):
Thanks for making the changes! I just updated the corresponding frontend PR (https://github.com/ShelterTechSF/askdarcel-web/pull/889) with the latest API changes, and I can confirm that adding addresses works (and editing still works).

It looks like you have a merge conflict in the Postman file, so you'll need to resolve those merge conflicts. In addition, I believe you are probably not passing the lint checks, since I see a few things in the diff that are not likely to pass lint, so you'll want to fix those up.

This is looking pretty good, and I think we're in good shape for getting this in by this weekend!

postman/Sheltertech_Data.postman_collection.json
```diff
@@ -1252,13 +1243,332 @@
 					"body": "{\n    \"eligibilities\": [\n        {\n            \"name\": \"Abuse or Neglect Survivors\",\n            \"id\": 1068,\n            \"feature_rank\": null\n        },\n        {\n            \"name\": \"Active Duty\",\n            \"id\": 1009,\n            \"feature_rank\": null\n        },\n        {\n            \"name\": \"Adoption\",\n            \"id\": 1,\n            \"feature_rank\": null\n        },\n        {\n            \"name\": \"Adults (31-54 years old)\",\n            \"id\": 1003,\n            \"feature_rank\": null\n        }\n    ]\n}"
 				}
 			]
+		},
+		{
+			"name": "Get Services By Categories",
```
richardxia, line 1248: I think this might have been committed by accident? I don't think this test is related to the changes in your PR, so it might have come from a merge conflict? If you don't have any intentional changes to this file for this PR, could you restore this back to what it is on `master`? The following commands should let you do that:

```sh
git fetch
git checkout origin/master -- postman/Sheltertech_Data.postman_collection.json
git add postman/Sheltertech_Data.postman_collection.json
git commit
```

app/controllers/change_requests_controller.rb
richardxia: If this comment isn't needed anymore, could you remove it?

app/controllers/change_requests_controller.rb
```diff
@@ -233,7 +305,12 @@ def get_field_change_hash(change_request)
   end
 
   def field_changes
-    params[:change_request].to_unsafe_h.map do |name, value|
+    if params[:change_request][:field_changes]
+     fields = params[:change_request][:field_changes]
```
richardxia, line 309:
```suggestion
      fields = params[:change_request][:field_changes]
```

Could you indent this by one more space to make it consistent with everything else?

Also, I'm honestly pretty surprised that Rubocop doesn't complain about this, given how much it complains about other stuff...

app/controllers/change_requests_controller.rb
```diff
@@ -53,6 +55,79 @@ def create_change_request
...
+      # Performing geoencoding here because we want to save the address before we hit persist_change
+      begin
+        a = Geokit::Geocoders::GoogleGeocoder.geocode address.address_1 + "," + address.city + "," + address.state_province
+        address.latitude = a.latitude
+        address.longitude = a.longitude
+
+      rescue => error
+        puts "google geocoding failed for address " + address.id.to_s + ": " + error.message
+
+      end
```
richardxia, line 102: Since this block of code is mostly the same as the other instance of it in this file, could you refactor it into a separate method that both locations call? I think the one part that is different is that in this block, you directly modify the `Address` object, but in the other block, it directly modifies the `field_changes` hash, so instead, I think you can just return the output of `GoogleGeocoder.geocode`.

richardxia (commented):
This looks pretty good, just have a few more small suggestions, but once those are addressed, I think this is good to merge!

richardxia (approved):
LGTM! Thanks for working on this!

PR #572, opened 2021-03-22
Update rails to 5.2.4.5

lgarofalo:


richardxia (approved):
LGTM, although in the future, we could try to get rid of Uglifier, since Webpack in the askdarcel-web repo should take care of all the JS/CSS post-processing.

PR #575, opened 2021-04-02
Remove uglifier dependency

richardxia:
Fixes issue in development environments introduced by https://github.com/ShelterTechSF/askdarcel-api/pull/572.

Adding the hard dependency on uglifier causes the Docker-based development environment to fail due to not having Node.js installed:

```
ExecJS::RuntimeUnavailable: Could not find a JavaScript runtime. See https://github.com/rails/execjs for a list of available runtimes.
/
```

Since there is not a single line of JavaScript code in this repository, I've removed the uglifier dependency as well as the entire Rails [asset pipeline](https://guides.rubyonrails.org/asset_pipeline.html), which is useless in a API-only Rails application.

I tested this locally by rebuilding the Docker image and confirming that `rake db:migrate` and the usual `docker-compose up api` commands work. I also tested that the frontend app can still successfully communicate with the API.


katerina-kossler (approved):
lgtm; was able to verify everything works in docker with these changes. Ran test suite and only the texting part of the suite had issues (I believe this was happening already).

richardxia:
Yep, looks like restarting the failed build worked. Thanks for confirming that this fixes the issue for you!

PR #576, opened 2021-04-13
feature: Delete Address via Change Request

katerina-kossler:
Add support to remove an existing address via a Post request to addresses/:address_id/change_request with the change_request field 'action' set to 'remove'.

### DIFF
--- a/app/controllers/change_requests_controller.rb
+++ b/app/controllers/change_requests_controller.rb
@@ -230,7 +230,6 @@ class ChangeRequestsController < ApplicationController
     object_id = change_request.object_id
     puts object_id
     field_change_hash = get_field_change_hash change_request
-
     if change_request.is_a? ServiceChangeRequest
       puts "ServiceChangeRequest"
       service = Service.find(change_request.object_id)
@@ -256,14 +255,19 @@ class ChangeRequestsController < ApplicationController
     elsif change_request.is_a? AddressChangeRequest
       puts "AddressChangeRequest"
       address = Address.find(change_request.object_id)
-      
-      a = geocode_address address
-      unless a.nil?
-        field_change_hash["latitude"] = a.latitude
+      if address && field_change_hash["action"] == "remove"
+        address.delete
+      else
+        a = geocode_address address
+        unless a.nil?
+          field_change_hash["latitude"] = a.latitude
+          field_change_hash["longitude"] = a.longitude 
         field_change_hash["longitude"] = a.longitude 
-      end
+          field_change_hash["longitude"] = a.longitude 
+        end
 
-      address.update field_change_hash
+        address.update field_change_hash
+      end
     else
       puts "invalid request"
     end
--- a/app/controllers/change_requests_controller.rb
+++ b/app/controllers/change_requests_controller.rb
@@ -18,7 +18,12 @@ class ChangeRequestsController < ApplicationController
     elsif params[:service_id]
       @change_request = ServiceChangeRequest.create(object_id: params[:service_id], resource_id: Service.find(params[:service_id]).resource_id)
     elsif params[:address_id]
-      @change_request = AddressChangeRequest.create(object_id: params[:address_id], resource_id: Address.find(params[:address_id]).resource_id)
+      if params[:change_request][:action]
+        action_code = params[:change_request][:action] !~ /\D/ ? params[:change_request][:action].to_i : params[:change_request][:action]
+        @change_request = AddressChangeRequest.create(action: action_code, object_id: params[:address_id], resource_id: Address.find(params[:address_id]).resource_id)
+      else
+        @change_request = AddressChangeRequest.create(object_id: params[:address_id], resource_id: Address.find(params[:address_id]).resource_id)
+      end
     elsif params[:phone_id]
       resource_id = Phone.find(params[:phone_id]).resource_id
       @change_request = PhoneChangeRequest.create(object_id: params[:phone_id], resource_id: resource_id)
@@ -255,17 +260,14 @@ class ChangeRequestsController < ApplicationController
     elsif change_request.is_a? AddressChangeRequest
       puts "AddressChangeRequest"
       address = Address.find(change_request.object_id)
-      if address && field_change_hash["action"] == "remove"
+      if change_request.remove?
         address.delete
       else
         a = geocode_address address
         unless a.nil?
           field_change_hash["latitude"] = a.latitude
           field_change_hash["longitude"] = a.longitude 
-        field_change_hash["longitude"] = a.longitude 
-          field_change_hash["longitude"] = a.longitude 
         end
-
         address.update field_change_hash
       end
     else
@@ -324,6 +326,9 @@ class ChangeRequestsController < ApplicationController
       elsif name == "eligibilities"
         field_change_hash[:field_name] = "eligibility_ids"
         field_change_hash[:field_value] = value.map { |c| c[:id] }.to_json.to_s
+      elsif name == "action"
+        field_change_hash[:field_name] = name
+        field_change_hash[:field_value] = @change_request.action
       else
         field_change_hash[:field_name] = name
         field_change_hash[:field_value] = value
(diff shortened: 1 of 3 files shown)


app/controllers/change_requests_controller.rb
```diff
@@ -256,14 +255,19 @@ def persist_change(change_request)
     elsif change_request.is_a? AddressChangeRequest
       puts "AddressChangeRequest"
       address = Address.find(change_request.object_id)
-      
-      a = geocode_address address
-      unless a.nil?
-        field_change_hash["latitude"] = a.latitude
+      if address && field_change_hash["action"] == "remove"
```
richardxia, line 258: The `action` is actually a [property of the ChangeRequest object itself](https://github.com/ShelterTechSF/askdarcel-api/blob/e8fa0c49a3f224f88668313a03200a48e0c896f9/app/models/change_request.rb#L5), not the actual database object it is meant to modify. I think we should grab the `action` from the ChangeRequest object, not the field changes hash, since the latter is meant to model the fields on the actual object being modified. Since `action` is a [Rails enum field](https://api.rubyonrails.org/v5.2.5/classes/ActiveRecord/Enum.html), you should be able to write code that looks like `change_request.remove?` to check the value of that field, although this is predicated on us actually properly populating that field on the ChangeRequest object in the first place.

Also, it looks like the in the frontend code, I am sending the stringified numeric values associated with each action, not the string name nor a normal number. [There is a comment that explicitly documents this](https://github.com/ShelterTechSF/askdarcel-web/blob/c5bf7ac26ec184e7ba0f7ac701103b1f4805986f/app/pages/OrganizationEditPage.jsx#L21-L25), although I am not certain whether those come from an earlier implementation of the multiple addresses feature that never got merged. It's strange because the Postman tests always send strings names, but the frontend code always sends stringified numbers, which suggests that this field was never really being checked properly before?

app/controllers/change_requests_controller.rb
```diff
@@ -256,14 +255,19 @@ def persist_change(change_request)
...
+        address.delete
+      else
+        a = geocode_address address
+        unless a.nil?
+          field_change_hash["latitude"] = a.latitude
+          field_change_hash["longitude"] = a.longitude 
         field_change_hash["longitude"] = a.longitude 
-      end
+          field_change_hash["longitude"] = a.longitude 
+        end
```
richardxia, line 267: Something looks very wrong about this code, since the indentation doesn't match the actual nesting of expressions, and it seems like we're assigning to the lat/long twice in a row. Perhaps it was a strange git merge conflict? Could you fix this up?

config/routes.rb
```diff
@@ -46,7 +46,7 @@
     resources :change_requests, only: :create
   end
   resources :addresses do
-    resources :change_requests, only: :create
+    resources :change_requests, only: %i[create destroy]
```
richardxia, line 49: It doesn't seem like we need to expose an actual `destroy` route, since to delete a resource, we issue the creation of a change request that itself represents the requested deletion.

postman/AskDarcel%20API.postman_collection.json
```diff
@@ -1,8 +1,8 @@
 {
 	"info": {
-		"_postman_id": "1f00be4c-4f11-4604-be5c-9548c7b31ab4",
+		"_postman_id": "82560b98-a655-4d4e-9b11-b784f4de9583",
 		"name": "AskDarcel API",
-		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
+		"schema": "https://schema.getpostman.com/json/collection/v2.0.0/collection.json"
```
richardxia, line 5: I think you accidentally reverted the version of the Postman JSON file schema, which is causing a large number of unrelated changed lines in the diff and consequently makes it more difficult to review the changes. Could you make sure your Postman is up to date and regenerate this file without all of the deleted lines?

richardxia (commented):
Thanks for working on this! This mostly looks like it's the right direction, but when I tried deleting an address from the askdarcel-web app, the askdarcel-api app rejected the API request. Although we don't necessarily have to completely match what the web app is sending right now, if we want to change how the frontend is formatting the `action` field, we'll need to test it across the different resource types being edited.

app/controllers/change_requests_controller.rb
katerina-kossler: ya not sure how this happened; reverting!

app/controllers/change_requests_controller.rb
katerina-kossler: I was able to check `change_request.remove?` however I am having difficulty implementing the action specification as anything other than a field change request. Am I missing something there?

richardxia: Oh weird, I took a look at the code a little bit more closely, and I saw these lines:

https://github.com/ShelterTechSF/askdarcel-api/blob/e8fa0c49a3f224f88668313a03200a48e0c896f9/app/controllers/change_requests_controller.rb#L309-L313

It looks like we have some logic to "fall back" to just `params[:change_request]` if `params[:change_request][:field_changes]` doesn't exist, which IMO is kind of confusing. But it's not worth changing this right now, so you can probably disregard parts of my suggestion here.

I think the main drawback to this "fallback" logic is that you can't actually read from the `action` when there are `field_changes` present, but in the event where we want to `remove` something, we won't have any `field_changes` anyway, so it won't matter. This might explain, though, why there's a discrepancy between the backend expecting strings and the frontend sending numerals for the action. Knowing this now, I'm inclined to change the frontend code to send strings for the action, since it just looks like we were never reading from the `action` in the first place before.

app/controllers/change_requests_controller.rb
```diff
@@ -18,7 +18,17 @@ def create_change_request
     elsif params[:service_id]
       @change_request = ServiceChangeRequest.create(object_id: params[:service_id], resource_id: Service.find(params[:service_id]).resource_id)
     elsif params[:address_id]
-      @change_request = AddressChangeRequest.create(object_id: params[:address_id], resource_id: Address.find(params[:address_id]).resource_id)
+      if params[:change_request][:action]
+        action_code = params[:change_request][:action]
+        if ['remove'].include?(action_code)
```
richardxia, line 23: Is this just the same as `if action_code == 'remove'`? My assumption would be that maybe you were originally trying to match against multiple different action_code strings but then later deleted the other strings, leaving just a single-element array here.

Could you write this in the more usual `action_code == 'remove'` style?

app/controllers/change_requests_controller.rb
```diff
@@ -256,14 +265,16 @@ def persist_change(change_request)
     elsif change_request.is_a? AddressChangeRequest
       puts "AddressChangeRequest"
       address = Address.find(change_request.object_id)
-      
-      a = geocode_address address
-      unless a.nil?
-        field_change_hash["latitude"] = a.latitude
-        field_change_hash["longitude"] = a.longitude 
+      if change_request.remove?
+        address.delete
```
richardxia, line 269: Generally, you'll want to call `#destroy`, not `#delete`. The difference is that `#destroy` will also delete any child models and call any other custom callback methods. See https://stackoverflow.com/a/22757533 for a slightly more detailed description and links to the official documentation.

Here, I believe the difference will become apparent with the Addresses–Services join table, where if you merely try to `#delete` an Address that has a relationship to a Service, then it will (probably?) cause a SQL constraints failure, because it won't try to destroy the Address–Service relationship, but the Address-Service relationship will have a foreign key to the Address you are trying to delete. I believe that `#destroy` will delete both the Address and its specific row in the Address-Service table, which will not hit an error.

It's been many years since I've seriously written Rails code like this, so there might be some mistakes in my description above, but I believe high level ideas are still correct.

richardxia (approved):
Two small comments, but I think this otherwise LGTM! Feel free to merge once addressing the final couple of things.

app/controllers/change_requests_controller.rb
katerina-kossler: @richardxia yes and yes!

app/controllers/change_requests_controller.rb
katerina-kossler: was not aware of this discrepancy!

PR #577, opened 2021-04-17
Feature: Expose new addresses endpoint for services

katerina-kossler:
Changes:
- Add route at /services/:service_id/addresses?address_id=:address_id to associate a service and address
- Post requires no params in the body
- Post fails if address or service do not exist 
- Added postman tests

Concerns:
- currently can associate a service to an address multiple times

### DIFF
--- a/app/presenters/services_presenter.rb
+++ b/app/presenters/services_presenter.rb
@@ -22,6 +22,6 @@ class ServicesPresenter < Jsonite
   property :schedule, with: SchedulesPresenter
   property :notes, with: NotesPresenter
   property :categories, with: CategoryPresenter
-  # property :addresses, with: AddressPresenter
+  property :addresses, with: AddressPresenter
   property :eligibilities, with: EligibilityPresenter
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -36,6 +36,7 @@ Rails.application.routes.draw do
     post :approve
     post :reject
     post :certify
+    post :addresses
     collection do
       get :featured
       get :pending
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -129,6 +129,18 @@ class ServicesController < ApplicationController
     render json: Service.all.count
   end
 
+  def addresses
+    if params[:service_id] && params[:address_id]
+      service = Service.find params[:service_id]
+      address = Address.find params[:address_id]
+      service.addresses << address
+      service.save!
+      render status: :created, json: { addresses: service.addresses}
+    else
+      render status: :precondition_failed
+    end
+  end
+
   private
 
   def remove_from_algolia(service)
--- a/postman/AskDarcel%20API.postman_collection.json
+++ b/postman/AskDarcel%20API.postman_collection.json
@@ -1,6 +1,6 @@
 {
 	"info": {
-		"_postman_id": "1f00be4c-4f11-4604-be5c-9548c7b31ab4",
+		"_postman_id": "04d60024-7618-4f3a-9257-15c0b6ff3958",
 		"name": "AskDarcel API",
 		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
 	},
@@ -181,7 +181,6 @@
 				{
 					"listen": "test",
 					"script": {
-						"id": "27b25cdd-5e05-4709-9fac-94a0e15c53a7",
 						"exec": [
 							"tests[\"Response time is less than 200ms\"] = responseTime < 2000;",
 							"",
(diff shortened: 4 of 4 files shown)


app/presenters/services_presenter.rb
```diff
@@ -22,6 +22,6 @@ class ServicesPresenter < Jsonite
   property :schedule, with: SchedulesPresenter
   property :notes, with: NotesPresenter
   property :categories, with: CategoryPresenter
-  # property :addresses, with: AddressPresenter
+  property :addresses, with: AddressPresenter
```
richardxia, line 25: It doesn't seem like the rest of your implementation required this to be uncommented. If that's correct, then could you leave it commented out? I'm not sure why it was commented out in the first place, but it's probably better to leave it commented out if we don't need it and if we're not sure why it was removed.

richardxia (commented):
> Add route at /services/:service_id/addresses?address_id=:address_id to associate a service and address

Could you change the query parameter to a path parameter? That is, have the address ID be specified as just another `/` component in the URL as opposed to something in the `?` section of the URL. Traditionally, query parameters were reserved for GET requests and data for a POST requests go in the body of a request.

> currently can associate a service to an address multiple times

Do you know if this causes multiple entries to be inserted into the database, or does it fail to save? @jjfreund, we should probably have a DB uniqueness constraint on the FK columns, or at least attempt to enforce that on the Rails side. As a pure Ruby fallback, we could at least try looking for the existence of the relationship before inserting another one.

app/presenters/services_presenter.rb
katerina-kossler: I was thinking that It's probably needed if we want to present a service with addresses but I'll leave it commented out for now.

app/controllers/addresses_controller.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class AddressesController < ApplicationController
+  def update
+    service = Service.find(params[:service_id])
+    address = Address.find(params[:id])
```
richardxia, line 6: Since these will raise an exception if either the Service or the Address don't exist, you should try to gracefully handle those cases instead of letting it cause an unhandled exception on the server side. Letting unhandled exceptions bubble up will result in a 500 error, which indicates that something unexpected happened on the server side. Not finding a Service or an Address is not an unexpected scenario, though, so instead we should handle that case and return some kind of 4xx-level error to indicate that the client passed invalid IDs.

See https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find for the documentation that describes the error-raising behavior.

app/controllers/addresses_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    if !service.addresses.include?(address)
```
richardxia, line 8: Nit: Normally, if I have both the `if` and the `else` parts of an `if` statement, I prefer to remove any negations in the condition by swapping the `if` and `else` bodies. It's just a minor aesthetic thing in order to avoid double negatives.

app/controllers/addresses_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+      service.addresses << address
+      service.save!
+      render status: :created
+      return
+    else
+      render status: :precondition_failed
```
richardxia, line 14: Thinking about this more, we probably don't want to return a `precondition_failed` in the case where the relationship already exists. This is especially the case if we're going to use the PUT method for this API call, since you generally want [PUT requests to be idempotent](https://www.infoq.com/news/2013/04/idempotent/).

([Idempotence](https://en.wikipedia.org/wiki/Idempotence#Computer_science_meaning) is a fancy word for "If I call this function multiple times with the same arguments, it should have the same effect as calling it only once," or in practice, "If I accidentally call this a bunch of times with the same argument, it won't do anything different than if I just called it once like I had intended to.")

I think if the relationship exists already, then we can return successfully, though we should just return a 200 (OK) or a 204 (No Content) instead of a 201 (Created), per the [specification for PUT](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.6).

app/controllers/addresses_controller.rb
katerina-kossler: @richardxia Great advice! I'll just use the find_by_id methods here to not throw errors and just make sure both are truthy (otherwise rendering a bad_request).

config/routes.rb
```diff
@@ -33,7 +33,7 @@
   resources :services do
     resources :change_requests, only: :create
     resources :notes, only: :create
-    resources :feedbacks, only: %i[create index]
```
richardxia, line 36: Looks like you accidentally reverted the feedback routes here. Could you restore this line?

richardxia (approved):
LGTM, once you restore the feedback route line!

PR #578, opened 2021-04-25
Update rails defaults for DB connection in dev

lgarofalo:
This change just exposes the postgres DB through localhost and updates rails to be able to find it if you don't have the other DB env variabbles set.

This allows me to do stuff like 
`bundle exec rails swagger` from my host and it uses the dockerized postgres DB to run.


richardxia (approved):
LGTM!

PR #581, opened 2021-04-30
Remove Association between Address and Service

katerina-kossler:
Added a #destroy method to the Addresses Controller.
Followed convention with other DELETE requests in the Api.

### DIFF
--- a/app/controllers/addresses_controller.rb
+++ b/app/controllers/addresses_controller.rb
@@ -1,21 +1,47 @@
 # frozen_string_literal: true
 
-# rubocop:disable Metrics/MethodLength
 class AddressesController < ApplicationController
   def update
     service = Service.find_by_id(params[:service_id])
     address = Address.find_by_id(params[:id])
     if service && address
-      if service.addresses.include?(address)
-        render status: :ok
-      else
-        service.addresses << address
-        service.save!
-        render status: :created
-      end
+      add_address_to_service(service, address)
+      return
     else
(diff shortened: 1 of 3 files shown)


app/controllers/addresses_controller.rb
```diff
@@ -1,21 +1,47 @@
 # frozen_string_literal: true
 
-# rubocop:disable Metrics/MethodLength
```
richardxia, line 3: Honestly, I feel that 5 lines per method is kind of insane. Our ChangeRequestsController definitely has some methods that are probably too long, but I really don't think breaking up these methods in AddressesController is helping with readability at all. I'd say feel free to leave this rubocop disable line in, and move the private methods back into the main methods.

We should probably separately configure rubocop to have a more reasonable maximum method length.

app/controllers/addresses_controller.rb
```diff
@@ -1,21 +1,47 @@
...
+      service.addresses << address
+      service.save!
+      render status: :created
+    end
+  end
+
+  def remove_address_from_service(service, address)
+    if service.addresses.include?(address)
+      service.addresses.delete(address)
+      service.save!
```
richardxia, line 41: Note that it's probably not necessary ta call save here, since services and addresses are a many-to-many relationship, which means that deleting the relationship corresponds to deleting a row in the join table, but without actually changing either model.

richardxia (approved):
LGTM!

PR #582, opened 2021-05-09
Add endpoint to render a service dump by site id

drcaramelsyrup:
SFFamilies is requesting an export of the current state of their services. We can give them a list of services, but it's ideal to include information such as the resource addresses, descriptions, schedules, etc. in addition so they can understand that the data is presented together.

Since our DB is normalized and stored in whatever data format is convenient, I'd say the easiest way to make a human readable spreadsheet like they expect is via a data dump endpoint like we do with Benetech, i.e. `/resources?category_id=all`. Much like with that endpoint, I think this may only be queried very infrequently (probably won't even mention it). It does make such data exports much easier, however, and we can prettify / linearize the json output into a csv format separately.

Sadly production has deviated a bit from staging data, so this will need to be pushed to production before I can give the SFFamilies team the data they want from it.


app/controllers/services_controller.rb
```diff
@@ -221,8 +233,7 @@ def find_services_eager_load_resources
       .includes(
         resource: [
           :addresses, :phones, :categories, :notes,
-          schedule: :schedule_days,
-          ratings: [:review]
```
drcaramelsyrup, line 225: It seems on the current staging DB including ratings results in an error, even with a query such as `/services?category_id=1,2,3` that should have worked before this PR. I think we dropped this table last year, anyway.

app/controllers/services_controller.rb
```diff
@@ -13,20 +13,32 @@ class ServicesController < ApplicationController
...
-    render json: ServicesWithResourceLitePresenter.present(all_services)
+    render json: ServicesWithResourceLitePresenter.present(matching_services)
   end
 
   def read_site_id_string
     @site_id_string = params[:site_id] || Site.find_by(site_code: 'sfsg').id.to_s
   end
 
+  def find_services(category_id, eligibility_id, site_id)
+    if category_id == 'all' || eligibility_id == 'all'
```
jjfreund, line 30: Wouldn't it be preferable to assume that you are looking for all categories/eligibilities if only site id is specified? I think the default assumption is that only the criteria explicitly passed in to the query string are used to filter results.

app/controllers/services_controller.rb
```diff
@@ -13,20 +13,32 @@ class ServicesController < ApplicationController
...
+      find_all(site_id)
+    # TODO: be able to parse both categories and eligibilities at once
+    elsif category_id
+      find_by_category(category_id, site_id)
+    elsif eligibility_id
+      find_by_eligibility(eligibility_id, site_id)
+    else
+      []
```
jjfreund, line 38: Should we return an error code if expected data is not being passed in?

jjfreund (changes requested):
It might be a good idea to modify the postman tests to ensure that these finders are actually under test. The current postman tests purporting to test the finds by category and eligibility aren't actually doing that.

app/controllers/services_controller.rb
drcaramelsyrup: Updated

drcaramelsyrup:
Also duplicated / expanded tests from Sheltertech_Data readonly api to include in our internal CI tests.

jjfreund (commented):
lgtm

PR #585, opened 2021-05-26
Fix service endpoint returning 500s (missing rating table)

drcaramelsyrup:
The SFFamilies service endpoint is currently serving 500s due to a missing ratings table expected by the services with resource presenter.

(Postman CI tests didn't catch this cause the existing tests in the AskDarcel API section for get services by categories are bogus, the ShelterTech_Data readonly api endpoint tests need to get copied over or also ran in CI.)


jjfreund (approved):
lgtm

PR #586, opened 2021-05-27
Remove Country from Addresses

katerina-kossler:
- Remove country column from addresses table (users will no longer enter a country on the front end - we will use 'USA' where needed)
- Remove country from specs and factories
- Update default country to 'USA' where needed


richardxia (approved):
Sorry for the delay, but this LGTM!

@jjfreund, what do you think about this change?

PR #587, opened 2021-06-08
Partial revert of 8f1a575c9b11f987578342bfbbe02063373863e7

richardxia:
For some reason we commented out addresses in the ServicesPresenter in 8f1a575c9b11f987578342bfbbe02063373863e7. This prevents the multiple addresses for services feature from working, so I am partially reverting that commit.


katerina-kossler (approved):
lgtm, I tried to uncomment this out at some point and didn't have any local issues so this should be good!

richardxia:
All the CI/deployment stuff is fixed now, so I'm finally merging this now.

PR #588, opened 2021-06-19
Add basic support for Github Actions

lexholden:


lexholden:
@jjfreund @richardxia @lgarofalo this just needs the secret for the `TEXTELLENT_AUTH_CODE` to be added here https://github.com/ShelterTechSF/askdarcel-api/settings/secrets/actions and it the tests should pass, then we can merge this to replace Travis.

Still need a task to build/push the latest docker image on master, easy enough to add, will do separately.

lgarofalo (approved):

lgarofalo:
I'll take care of `TEXTELLENT_AUTH_CODE` and update here when done.

lgarofalo:
@lexholden Added PR to wire the secret env var from that link you sent into the build environment.

.github/workflows/images.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: docker/setup-qemu-action@v1
+      - uses: docker/setup-buildx-action@v1
+      - uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - uses: docker/build-push-action@v2
```
lgarofalo, line 23: I think the step to tag the image is still missing, or is that done elsewhere?

lgarofalo: Ugh, I don't know why I didn't see, right below here it does this 🤦‍♂️

lexholden: Pretty sure this is all handled in this one action, you can see what this pushed to my personal docker hub here: https://hub.docker.com/repository/docker/lexholden/askdarcel-api

There is no version tag, it just pushes `latest`. You're right, we would probably need to support also running this on a release branch with the release name as a tag. Not sure how to do this, can find some time this weekend to figure it out if you think it's a blocker to merging this?

lgarofalo: https://github.com/docker/build-push-action/blob/master/docs/advanced/tags-labels.md

Looks like a good example for this

PR #590, opened 2021-06-20
Update dev Dockerfile to stretch

lgarofalo:


jjfreund (approved):

Dockerfile.dev
```diff
@@ -1,6 +1,6 @@
 FROM ruby:2.7.2
 
-RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ jessie-pgdg main' > /etc/apt/sources.list.d/pgdg.list && \
+RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main' > /etc/apt/sources.list.d/pgdg.list && \
```
richardxia, line 3: Wait, is this even right version? My local Docker image is on Debian version 10, which is buster, not stretch.

lgarofalo: That's a good question, I'll investigate, but this did at least fix the build.

PR #597, opened 2021-06-30
Add algolila property for holding associated sites to services and re…

wi11:
We want to add this attribute to both resources and services so that SF Families can query using this attribute.


jjfreund (approved):
lgtm!

PR #600, opened 2021-08-12
Updates to api doc generation and tests

lgarofalo:


lgarofalo:
With this PR, then do `bundle exec rails swagger` and it updates the api doc swagger file.

swagger/v1/swagger.json
```diff
@@ -61,13 +59,14 @@
...
                   {
+                    "updated_at": "2021-08-12T03:11:24.138Z",
                     "alternate_name": null,
                     "certified": false,
                     "email": null,
                     "id": 68,
                     "legal_status": null,
                     "long_description": null,
-                    "name": "Mosciski-Bode",
+                    "name": "Greenfelder and Sons",
```
richardxia, line 69: We might want to make the data fixtures deterministic for the Swagger stuff, or else every time we regenerate them, we're going to get a ton of differences.

spec/requests/eligibilities_spec.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
+require 'swagger_helper'
```
richardxia, line 4: I think you can just put `require 'swagger_helper`` inside of either `rails_helper.rb` or `spec_helper.rb` and it'll automatically get required. That way, you won't have to put `require 'swagger_helper'` in all of these spec files. Or alternatively, you could just replace the `require 'rails_helper'` calls because `swagger_helper.rb` already requires `rails_helper`.

PR #601, opened 2021-09-09
Use algolia search API to get site results instead of postgres

wi11:
Currently WIP


drcaramelsyrup:
I think before merging we said all we wanted was adding some form of unittests, right @jjfreund ?

- [ ] postman unit tests

PR #608, opened 2021-11-21
Fix Algolia services search filters, add free text search and exclusive tag matching

drcaramelsyrup:
This PR:
- fixes some issues with the Algolia services search currently in staging: a) fixing a crash if the category / eligibility id requested is not present in the database, and b) replaced `filter_map` with `select` then `map`, as `filter_map` is not accessible in the Ruby version in staging
- adds a `text` query param to the endpoint to add Algolia free text filtering as well
- adds a `match_all_tags` param to use ANDs instead of ORs when matching multiple tags at once

the latter two are eagerly awaited features from the SFFamilies team.

### DIFF
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -152,17 +152,17 @@ class ServicesController < ApplicationController
   end
 
   def eligibilities_filter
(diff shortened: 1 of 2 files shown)


jjfreund (approved):
lgtm, we should try to figure out a better way to set up test data (rspec migration would do it) to more reliably test the free text search, and other things.

PR #614, opened 2022-05-04
Update Postgres Client package to 13. 9.5 Doesnt seem to exist anymor…

schroerbrian:
…e. Replace deprecated algolia index method batch_snynoyms. Update staging DB hostname.

I set up the API environment on my new comp and ran into some of these issues while creating the Docker build and importing the staging DB. Looks like we're quite behind with our Postgres Client version. We're also using deprecated Algolia index methods. We also need to do a deep clone of the local db config object (as `.clone` just creates a pointer to the hash reference, which Rails considers a frozen hash). And lastly, I had to update the hostname of the staging API to the correct IP as `staging-db.askdarcel.org` points to the wrong one.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -2,9 +2,10 @@ FROM ruby:2.7.4
 
 RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main' > /etc/apt/sources.list.d/pgdg.list && \
    wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - && \
-   apt-get update && \
-   apt-get install -y postgresql-client-9.5 && \
+   apt-get update && apt list -a postgresql-client && \
+   apt-get install -y postgresql-client-13 && \
    rm -rf /var/lib/apt/lists/*
+
 # throw errors if Gemfile has been modified since Gemfile.lock
 RUN bundle config --global frozen 1
 
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -2,10 +2,9 @@ FROM ruby:2.7.4
 
 RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main' > /etc/apt/sources.list.d/pgdg.list && \
    wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - && \
-   apt-get update && apt list -a postgresql-client && \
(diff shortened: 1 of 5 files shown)


richardxia (approved):
Wow, Postgres 9.5 is super old. Looks like it was first released in 2016, and it predates the new versioning scheme where they bump the first digit a lot more often.

We should probably also upgrade the actual Postgres server version, which would be more of an issue if it's out of date.

Can you also upgrade [the postgres version in the docker-compose file](https://github.com/ShelterTechSF/askdarcel-api/blob/8c0b890d18a4cba5ba1ea87f15b7654f7ff78c62/docker-compose.yml#L7)? That specifies the version of the Postgres server we use with local docker development.

schroerbrian:
> Wow, Postgres 9.5 is super old. Looks like it was first released in 2016, and it predates the new versioning scheme where they bump the first digit a lot more often.
> 
> We should probably also upgrade the actual Postgres server version, which would be more of an issue if it's out of date.
> 
> Can you also upgrade [the postgres version in the docker-compose file](https://github.com/ShelterTechSF/askdarcel-api/blob/8c0b890d18a4cba5ba1ea87f15b7654f7ff78c62/docker-compose.yml#L7)? That specifies the version of the Postgres server we use with local docker development.

I bumped the postgres version in the docker-compose file, but then got the following error in my API console, as well as error responses from the API in my browser: 

```askdarcel-api-api-1  | Failed to submit event: PG::ConnectionBad: could not translate host name "db" to address: Name or service not known
askdarcel-api-api-1  | ActiveRecord::ConnectionNotEstablished (could not translate host name "db" to address: Name or 
askdarcel-api-api-1  | ):```

I can do more debugging later, but I just moved it back for now while I focus on multi-location work. We can keep this PR open for the time being until we figure out what the issue is with this.

richardxia:
The `askdarcel-api-api-1` in that line implies that it's the `api` service that isn't able to talk to the `db` service. Is there are log output for the `db` service (I think it'd look like `askdarcel-api-db-1`)? It might be that you can't just have PostgreSQL 13 read the database files created by PostgreSQL 9.5, so you might need to totally purge the `db` service along with its persistent volume. I think you can do that with `docker-compose down --volumes`, which will remove all containers and volumes.

schroerbrian:
> docker-compose down --volumes

Yep, that did the trick! I logged the `db` service and got a message about PG 13 not being able to read files created by 9.5. Then I tore down the service/volume and rebuilt. Working fine now.

Thanks Richard!

PR #616, opened 2022-05-12
Deactivate services when deactivating resources.

richardxia:
It looks like we've had a very long-standing bug where a deactivated resource's services still show up in the Algolia search results. Turns out it's because we don't deactivate services when we deactivate resources, which keeps the services around in the Algolia search results.

This updates the `Resources#destroy` route, which is what the Deactivate button hits, to also deactivate associated services and remove them from Algolia.

This won't fix existing services in prod that need to be deactivated, so we'll probably need to search for all services that are `approved` but whose resource is not `approved` and mass deactivate them. We may also need to do a full Algolia reindex, but I believe we do that once a week, so it'll eventually correct itself.

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -33,12 +33,18 @@ class ResourcesController < ApplicationController
     render status: :ok
   end
 
-  def destroy
+  def destroy # rubocop:disable Metrics/MethodLength
     resource = Resource.find params[:id]
     if resource.approved?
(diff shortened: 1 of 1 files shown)


schroerbrian:
LGTM! Thanks for the clear description. I'll just go ahead and merge this. I'll make a note to deploy the API as well if @jjfreund and I are able to meet tomorrow

richardxia:
Thanks!

PR #624, opened 2022-06-02
Make Debian version match between Docker base image and PostgreSQL apt repository.

richardxia:
Having these be mismatched may cause packages to not be found. The
official Ruby images provide multiple versions of Debian for the same
Ruby version, so here we pin the Debian version to Buster (the latest
supported by Ruby 2.7) and make sure the PostgreSQL package matches.

This works for my Intel-based Mac, but someone with an M1 should confirm that this works for them as well so that we stop ping-ponging back and forth between breaking stuff on the other architecture.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,6 +1,6 @@
-FROM ruby:2.7.4
+FROM ruby:2.7.6-buster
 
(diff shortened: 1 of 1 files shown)


schroerbrian (approved):
I set up the API/DB images from scratch using this branch. It works on my M1 mac. Good work!

richardxia:
Awesome, thanks for testing this out locally!

PR #625, opened 2022-06-03
Add is_parent column and EligibilityRelationship table

wi11:


db/migrate/20220602025217_add_is_parent_to_eligibilities.rb
```diff
@@ -0,0 +1,5 @@
+class AddIsParentToEligibilities < ActiveRecord::Migration[6.1]
+  def change
+    add_column :eligibilities, :is_parent, :boolean, :default => false
```
jjfreund, line 3: In theory we don't need this -- you could query to see if an eligibility has any children and if so consider it a parent -- and data denormalization is generally not great but I think it should be fine in this case.

lib/sheltertech/db/fixture_populator.rb
```diff
@@ -580,6 +594,12 @@ module Constants # rubocop:disable Metrics/ModuleLength
         'Low-Income'
       ].freeze
 
+      PARENT_ELIGIBILITY_NAMES = [
+        'Foster Youth',
```
jjfreund, line 598: We might want to consider adding data that mimics UCSF's structure in a later PR.

jjfreund (approved):
Had a couple comments. Might be worth thinking about but lgtm.

PR #635, opened 2022-08-25
Add document model and join table with services

wi11:
Following PRs will contain populating the test DB fixture with some documents.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+class Document < ApplicationRecord
+end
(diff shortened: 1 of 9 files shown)


jjfreund (approved):
Looks good, so merging in to unblock Brian. Two things here:

1) Might be better to use faker to generate realistic data than using static test data, and

2) As it turns out really what Brian wants to display is the language associated with the document. For now he can just include that information in the name but it might be worth modeling languages in the long run -- either they could be an enum or a separate model that the document would be associated with. That sounds like a larger effort though.

PR #637, opened 2022-09-02
626 add instructions to services

jjfreund:


app/models/service.rb
```diff
@@ -22,6 +22,7 @@ class Service < ActiveRecord::Base
   accepts_nested_attributes_for :addresses
```
schroerbrian, line 22: I think we will want to add `accepts_nested_attributes_for :instructions` here. At least when creating new services (updating them is another story), it seems that our pattern is to include the attributes of any child models in the payload (as we do with addresses, schedules, and notes).

schroerbrian (changes requested):
LGTM! I have one comment regarding accepting nested instruction attributes when creating a service, but other than that I think this is good to go.

schroerbrian (approved):

app/models/service.rb
jjfreund: No harm in adding that, I guess. But aren't we going to persist these separately through dedicated endpoints?

schroerbrian: Yes – I can look into using the dedicated endpoints for both creating new instructions on services. But at the least adding the accepts nested attributes will hold us over in the meantime. 

I will definitely be using the dedicated endpoints for updating instructions on services (which sort of follows our current pattern except I don't think we will be using change requests for instructions updates).

PR #638, opened 2022-09-02
633 add long term housing categories

jjfreund:


db/sql/covid-long-term-housing.sql
```diff
@@ -0,0 +1,61 @@
...
+(1000011, 1100047);
+
+insert into category_relationships (parent_id, child_id) values
+(1100046, 1100048);
+
+insert into category_relationships (parent_id, child_id) values
+(1100046, 1100049);
+
+insert into category_relationships (parent_id, child_id) values
+(1100046, 1100045);
```
schroerbrian, line 46: `1100045` here should be `1100050` so as to prevent the error of an incorrect child subcategory that we noticed in Staging/Prod

schroerbrian (changes requested):
There is one error in the script where you're passing an incorrect child subcategory ID. Everything else is looking good.

PR #640, opened 2022-09-09
Include documents when fetching services

wi11:
Manually tested on local rails app, able to see the documents associated with a service as part of the services output:
http://localhost:3000/services/1 =>
`documents":[{"id":1,"url":"document_url"}]`

Note: The above output isn't the full output, i just grabbed the part which displays the documents.

### DIFF
--- a/app/presenters/document_presenter.rb
+++ b/app/presenters/document_presenter.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class DocumentPresenter < Jsonite
+  property :id
+  property :url
+  property :name
+  property :description
+end
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -218,7 +218,7 @@ class ServicesController < ApplicationController
   end
 
   def services
-    Service.includes(:notes, :categories, :eligibilities, :addresses, schedule: :schedule_days)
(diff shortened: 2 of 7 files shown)


app/presenters/document_presenter.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class DocumentPresenter < Jsonite
+  property :id
+  property :url
```
schroerbrian, line 5: Thanks Will! This is super helpful. 

There is one other property that we want, perhaps we can call it `description`. We'll be using it to list the language of the document. @jjfreund -- I believe we decided it would be a string right, rather than an enum?

wi11: If we use an enum, would it be better to name the field "language" or something like that? Description sounds like it applies to much more than just the language of the doc

schroerbrian: That is a good point. I would agree with that. I think we ultimately decided to go with a string and to call the field `description` so that the model could potentially be used to represent other sorts of data.

jjfreund: Document already has the following properties. They should be added to the presenter.

      t.string :name
      t.string :description

wi11: Updated, also added back the migration file that was accidentally deleted.

jjfreund (approved):

PR #644, opened 2022-09-19
627 create data model for breaking news section

james7770:


jjfreund (approved):
lgtm

app/models/news_article.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class NewsArticle < ActiveRecord::Base
+  belongs_to :service
```
jjfreund, line 4: it does not belong to service

PR #646, opened 2022-09-22
Add Instructions and Documents to Service indexing function.

schroerbrian:
Need these to ensure that `document` and `instruction` changes on a `service` are reflected in Algolia.

### DIFF
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -146,6 +146,16 @@ class Service < ActiveRecord::Base
         resource.sites.map(&:site_code)
       end
 
+      add_attribute :documents do
+        documents.map do |d|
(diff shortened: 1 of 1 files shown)


app/models/service.rb
```diff
@@ -146,6 +146,16 @@ class Service < ActiveRecord::Base
         resource.sites.map(&:site_code)
       end
 
+      add_attribute :documents do
+        documents.map do |d|
+          { id: d.id, url: d.url, description: d.description }
```
schroerbrian, line 151: The id is included so that the React frontend can use it to maintain a map between each specific document object and UI changes

wi11 (approved):
lgtm

PR #647, opened 2022-09-27
628 add crud endpoints for breaking news

james7770:
postman tests added


postman/AskDarcel%20API.postman_collection.json
```diff
@@ -2656,6 +2656,166 @@
...
 			"response": []
+		},
+		{
+			"name": "Add Breaking News",
+			"event": [
+				{
+					"listen": "test",
+					"script": {
+						"exec": [
+							"pm.test(\"Adding document\", function () {",
```
devbrianschroer, line 2667: I think we want to change "document" to "news article" or whatever on this line and in any instances below

devbrianschroer (approved):
Looks good. The only thing that I noticed is that in the `postman_collection.json` the script.exec uses "document" instead of "news article". I'm assuming this is because it was just copied and pasted, but I think it should be changed for clarity's sake.

schroerbrian (approved):
LGTM!

PR #648, opened 2022-09-28
642 crud endpoints for instructions

jjfreund:


schroerbrian (approved):
Looks great! I have only one comment which is to add the `id` property to the `Instructions_Presenter` so we can pass it when updating a given instruction.

PR #651, opened 2022-09-28
Use focal postgres package rather than xenial as xenial is no longer …

schroerbrian:
…supported.

Apparently we have been installing a Postgres package from an outdated version of linux for some time now. Postgres removed the package from their apt repo (https://wiki.postgresql.org/wiki/Apt) about a week ago, so our `publish-latest` builds have been failing since. This PR updates our Dockerfile to install the package from a newer repo (using Focal), which is the OS that our base Docker images are using. I tested the job [here](https://github.com/ShelterTechSF/askdarcel-api/actions/runs/3147092778/jobs/5116270033) by temporarily altering our `publish-latest` job to use a test tag, alongside the update to the Dockerfile


schroerbrian:
Just going to go ahead and merge this

PR #652, opened 2022-10-07
Add status to service presenter.

schroerbrian:
Including this prop so that the frontend can perform a redirect on service listing pages where the service is deactivated.


richardxia (approved):
LGTM!

PR #653, opened 2022-10-15
Only reindex resource or service on update if its status is 'approved'

schroerbrian:


app/models/resource.rb
```diff
@@ -38,7 +38,9 @@ class Resource < ActiveRecord::Base
 
     # Important: Use Resource.reindex! and Service.reindex! to reindex/create your index
     # rubocop:disable Metrics/BlockLength
-    algoliasearch index_name: "#{Rails.configuration.x.algolia.index_prefix}_services_search", id: :algolia_id do
+    algoliasearch if: :status_approved,
```
richardxia, line 41:
```suggestion
    algoliasearch if: :approved?,
```

As much as I don't like the magicalness of it, `:approved?` is already automatically defined to be the same as the `status_approved` method you created here, since Rails will automatically define methods for each possible enum value. Here is the documentation for this: https://api.rubyonrails.org/v5.1/classes/ActiveRecord/Enum.html

I haven't tested this, so we probably do want to run the indexing task at least once locally with this method, but I'm pretty sure it'll work.

richardxia (approved):
LGTM aside from my small comment on using the existing `:approved?` method. Thanks again for digging into this and noticing that saving a model would automatically add it back to the Algolia index!

PR #656, opened 2022-11-23
Add default lat & lng to services indexed by Algolia

tpetersen0308:
Services without addresses are not discoverable in search. This PR sets a default lat/lng on the Algolia geolocation when an address cannot be found for a service or resource.

The testing method is a little hacky. It looks like the Algolia setup runs before we seed the db. Short on time, this was what I came up with:

1. Paste the following into `/lib/sheltertech/db/fixture/populator.rb` at line 88:
`
        resource = FactoryBot.create(:resource,
                                     name: "Travis Test Org",
                                     short_description: "test",
                                     long_description: Faker::ShelterTech.description,
                                     website: "test.fake",
                                     categories: categories)
        resource.addresses = []
        resource.save
        service = FactoryBot.create(:service,
                                    resource: resource)
        service.save
`
2. run `docker-compose run --rm api rake db:setup db:populate`
3. run the api
4. run the client
5. search for "Travis Test Org"
6. expect search results like the following:
Screen Shot 2022-11-22 at 5 06 44 PM


schroerbrian:
Thanks so much for doing this. It has been sorely needed for some time!

Upon [further research](https://www.algolia.com/doc/guides/managing-results/refine-results/geolocation/#enabling-geo-search-by-adding-geolocation-data-to-records), I found that the `_geoloc` attribute seems to be an internal Algolia prop that determines whether a record is included in the search results or not, if results are returned by geolocation. The `addr.latitude` and `attr.longitude` props that you edited above are actually used by our code to plot the actual location on the map on our results page. You can take a look at the `SearchMap.tsx` file to see how this is done. 
(Though to further confuse things, our individual service page, like [this one](sfserviceguide.org/services/3578) uses the `_geoloc` prop, which we probably should not do.)

Anyhow this is just to say, I think we should therefore leave the `.longitude` and `.latitude` attrs alone and only set a default for the `_geloc` attribute. 

I'm going to tag @jjfreund and @richardxia here in case they have anything to add in terms of how the data and Algolia indexing interact. 

Let me know if you have any more questions and thanks again!!

tpetersen0308:
> Thanks so much for doing this. It has been sorely needed for some time!
> 
> Upon [further research](https://www.algolia.com/doc/guides/managing-results/refine-results/geolocation/#enabling-geo-search-by-adding-geolocation-data-to-records), I found that the `_geoloc` attribute seems to be an internal Algolia prop that determines whether a record is included in the search results or not, if results are returned by geolocation. The `addr.latitude` and `attr.longitude` props that you edited above are actually used by our code to plot the actual location on the map on our results page. You can take a look at the `SearchMap.tsx` file to see how this is done. (Though to further confuse things, our individual service page, like [this one](sfserviceguide.org/services/3578) uses the `_geoloc` prop, which we probably should not do.)
> 
> Anyhow this is just to say, I think we should therefore leave the `.longitude` and `.latitude` attrs alone and only set a default for the `_geloc` attribute.
> 
> I'm going to tag @jjfreund and @richardxia here in case they have anything to add in terms of how the data and Algolia indexing interact.
> 
> Let me know if you have any more questions and thanks again!!

Thanks @schroerbrian, I've addressed your feedback!

richardxia:
> The testing method is a little hacky. It looks like the Algolia setup runs before we seed the db. Short on time, this was what I came up with:

We actually have a Rake task for reindexing the Algolia index on all Resources and Services, though I'm now realizing that it's not documented in our README. It's defined in [lib/tasks/algolia.rake](https://github.com/ShelterTechSF/askdarcel-api/blob/0a6ce81f1a395bcc8da395d2c379d652e82764b0/lib/tasks/algolia.rake), and you should be able to run it via `docker-compose run --rm api bundle exec rake algolia:reindex`.

> I'm going to tag @jjfreund and @richardxia here in case they have anything to add in terms of how the data and Algolia indexing interact.

I don't think I have much to add here, but yeah, I like just changing the default for the Algolia indexing job rather than the actual DB field.

.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 3.0.4
```
richardxia, line 1: I think we probably don't want to commit this, unless we wanted to provide a different way of getting the app set up for development. At the very least, we probably want to make sure that we specify a consistent version of Ruby across all the different environments, and it looks like Dockerfile.dev uses Ruby 2.7.6.

tpetersen0308: Ah, whoops. This was committed accidentally

app/models/resource.rb
```diff
@@ -119,6 +119,7 @@ class Resource < ActiveRecord::Base
       end
     end
     # rubocop:enable Metrics/BlockLength
+    puts algoliasearch.inspect
```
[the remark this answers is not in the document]
tpetersen0308: 🤦‍♂️

app/models/resource.rb
```diff
@@ -45,10 +45,10 @@ class Resource < ActiveRecord::Base
       attributesForFaceting %i[categories open_times eligibilities associated_sites type]
       # Define attributes used to build an Algolia record
       add_attribute :_geoloc do
-        if addresses.blank?
-          nil
-        else
+        if addresses.any?
```
schroerbrian, line 48: I'm wondering if we want to replace `blank` with `any` (and `empty`) in the service.rb file below. Apparently, `blank` can be called on `nil` values and on any object actually; whereas `any` and `empty` will raise a `nomethod` exception if called on `nil`. A `resource` and `service` should always at least have an empty addresses array (I believe), but I think it might be better to be safe and stick with `blank`.

schroerbrian (approved):
Once comments are addressed, feel free to merge in. Thanks again for your help. This has been biting us in the ass for some time!

app/models/resource.rb
tpetersen0308: Ah I had changed it because `any?` gave me better results when I tested in `irb`. But it looks like `blank?` is an active record method or something, because it works the way you're describing when I test in the rails console. Happy to change it back.

richardxia: You can also use the [safe navigation operator](https://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/) (`&.`) to safely attempt to access methods on objects that might be `nil`. `addresses&.any?` will return `nil` if `addresses` is `nil` but will otherwise invoke `any?` on `addresses` when its not `nil`.

schroerbrian: I think using the `safe navigation operator` and leaving your changes would be a good way to go. Thanks for the info, Richard.

tpetersen0308:
@richardxia comments addressed, please let me know if this looks alright to you!

richardxia (approved):
Thanks, LGTM!

schroerbrian:
Going to go ahead and merge. Great work @tpetersen0308 !!

PR #663, opened 2023-01-27
1188 html to pdf endpoint

schroerbrian:
Adds html to pdf endpoint on the ServiceController. I will need to add the PDF Crowd API key to our Kubernetes secrets manager. I am still waiting to get a credit card so that I can sign up. Regardless, this code will work with at least the demo key.

### DIFF
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -69,7 +69,8 @@ class ServicesController < ApplicationController
       render(status: 500)
     end
   rescue Pdfcrowd::Error => e
-    render plain: e.getMessage, status: e.getCode
+    puts "Failed to convert HTML to PDF: #{e}"
+    render plain: "There was an error getting the PDF. Please try again", status: 500
   end
 
   def featured
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -55,14 +55,15 @@ class ServicesController < ApplicationController
   end
 
   def convert_to_pdf
-    client = Pdfcrowd::HtmlToPdfClient.new("demo", "ce544b6ea52a5621fb9d55f8b542d14d")
-    pdf = client.convertUrl(params[:url])
-
-    send_data pdf,
-              {
-                type: "application/pdf",
-                disposition: "attachment; filename*=UTF-8''#{ERB::Util.url_encode('result.pdf')} }"
-              }
+    if Rails.configuration.x.pdfcrowd.api_key
+      client = Pdfcrowd::HtmlToPdfClient.new("demo", Rails.configuration.x.pdfcrowd.api_key)
+      pdf = client.convertUrl(params[:url])
+      send_data pdf,
+                { type: "application/pdf",
+                  disposition: "attachment; filename*=UTF-8''#{ERB::Util.url_encode('result.pdf')} }"}
+    else
+      render(status: 500)
+    end
   rescue Pdfcrowd::Error => e
     render plain: e.getMessage, status: e.getCode
   end
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -65,8 +65,6 @@ class ServicesController < ApplicationController
       send_data pdf,
                 { type: "application/pdf",
                   disposition: "attachment; filename*=UTF-8''#{ERB::Util.url_encode('result.pdf')} }" }
-    else
-      render(status: 500)
     end
   rescue Pdfcrowd::Error => e
     puts "Failed to convert HTML to PDF: #{e}"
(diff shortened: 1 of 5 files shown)


app/controllers/services_controller.rb
```diff
@@ -54,6 +54,24 @@ def show
     render json: ServicesWithResourcePresenter.present(service)
   end
 
+  def init_pdf_crowd_client
+    Pdfcrowd::HtmlToPdfClient.new(Rails.configuration.x.pdfcrowd.username, Rails.configuration.x.pdfcrowd.api_key)
+  end
```
richardxia, line 59: I think this type of stuff is normally supposed to go into `config/initializers` so that you only do the initialization once on Rails startup, not for each request. Could you create a `pdfcrowd.rb` file with this in it? I think `algoliasearch.rb` and `sentry.rb` are good examples to borrow from.

app/controllers/services_controller.rb
```diff
@@ -54,6 +54,24 @@ def show
...
+      client = init_pdf_crowd_client
+      pdf = client.convertString(params[:html])
+      send_data pdf,
+                { type: "application/pdf",
+                  disposition: "attachment; filename*=UTF-8''#{ERB::Util.url_encode('result.pdf')} }" }
+    else
+      render(status: 500)
+    end
+  rescue Pdfcrowd::Error => e
+    render plain: e.getMessage, status: e.getCode
```
richardxia, line 72: Will this show up to end users? Maybe we should render a generic error message but log the more detailed error so that it shows up in Sentry.

config/routes.rb
```diff
@@ -36,9 +36,12 @@
...
     resources :addresses, only: %i[update destroy]
-    post :approve
-    post :reject
-    post :certify
+    collection do
+      post :approve
+      post :reject
+      post :certify
+      post :html_to_pdf
+    end
```
richardxia, line 44: Could you explain these changes here? The [Rails documentation](https://guides.rubyonrails.org/routing.html#adding-collection-routes) is kind of circular and says "To add a route to the collection, use a collection block", but it doesn't really define what a "collection" is. I think it means it defines the URL under `/services/approve` instead of `/services/:id/approve`, but I'm not 100% confident.

In either case, did you mean to change the routing configuration for the `approve`, `reject`, and `certify` routes? And could this `collection` block be merged with the `collection` block below it?

app/controllers/services_controller.rb
schroerbrian: I am having some trouble setting this one up. It seems like our other initializers like `algoliasearch.rb` and `sentry.rb` are setting config values for their respective module; whereas this line inits a new `Pdfcrowd::HtmlToPdfClient` class instance.. Is there some way I can then access this class instance in the `services_controller`? 🤔

config/routes.rb
schroerbrian: Oops, I misunderstood the collection block's purpose. I have just added the new endpoint to the existing block and moved the other routes out of the collection block.

app/controllers/services_controller.rb
schroerbrian: Sure, I've updated to return a generic errors msg. I just added a `puts` for logging. I couldn't find any other examples of logging across our codebase other than `puts`

app/controllers/services_controller.rb
richardxia: Sorry for the delayed reply, but does it not "just work" to instantiate the client in the initializer file? Were you hitting some kind of error when doing that?

app/controllers/services_controller.rb
richardxia: Hmm, I guess we're not really following best practices, but what I had in mind was using the [Rails logger](https://guides.rubyonrails.org/debugging_rails_applications.html#sending-messages).

app/controllers/services_controller.rb
schroerbrian: TIL that you have to assign the class instantiation to a capitalized var to make it globally accessible in Rails. I have a fix for this in a future PR. https://stackoverflow.com/questions/23588674/rails-how-to-set-global-instance-variables-in-initializers

richardxia: Ah, I see. You could also probably create a `module` (really just functioning as a namespace) and then an instance variable within that module's object (similar to a static variable in Java or a class variable in Python). See https://stackoverflow.com/a/8784166 for an example.

The minor benefit of creating an instance variable with access methods on a module vs. just assigning the thing to the capitalized initial letter variable is that the latter is actually syntax for a constant, and constants can never be reassigned to. Modules, on the other hand, can be reopened and their contents modified, which I think would be a better fit for this, since it would simplify cases such as unit testing where we may want to change the value.

PR #668, opened 2023-02-03
Algolia Rails client logs to stderr

drcaramelsyrup:
Attempt to address https://github.com/ShelterTechSF/askdarcel-api/issues/662 by configuring Algolia client to log to $stderr. Requires `algoliasearch-rails` version 2.2.0 (ran `bundle update algoliasearch-rails`) to set a custom logger (see https://github.com/algolia/algoliasearch-rails/pull/420).


richardxia (approved):
Sorry for the delay, but this LGTM! I'm not sure if there's an easy way to test this, since we'd have to trigger some internal error in the Algolia client library, but this looks correct to me.

PR #680, opened 2023-03-22
Add LD_PRELOAD ENV commands to Dockerfile to circumvent GRPC SegFault…

schroerbrian:
… bug

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,3 +20,6 @@ RUN mkdir -p /var/lib/dpkg/alternatives /var/lib/dpkg/info /var/lib/dpkg/parts /
   apt-get install -y libglib2.0-dev && \
   apt-get install -y postgresql-client-common && \
   rm -rf /var/lib/apt/lists/*
+
+ENV LD_PRELOAD=$LD_PRELOAD:/lib/x86_64-linux-gnu/libjemalloc.so.2\ bundle\ exec\ ruby\ server.rb
+ENV LD_PRELOAD=$LD_PRELOAD:/lib/x86_64-linux-gnu/libjemalloc.so.2\ bundle\ exec\ ruby\ client.rb


Dockerfile
```diff
@@ -19,4 +19,7 @@ RUN mkdir -p /var/lib/dpkg/alternatives /var/lib/dpkg/info /var/lib/dpkg/parts /
   apt-get update && \
   apt-get install -y libglib2.0-dev && \
   apt-get install -y postgresql-client-common && \
-  rm -rf /var/lib/apt/lists/*
\ No newline at end of file
+  rm -rf /var/lib/apt/lists/*
+
+ENV LD_PRELOAD=$LD_PRELOAD:/lib/x86_64-linux-gnu/libjemalloc.so.2\ bundle\ exec\ ruby\ server.rb
+ENV LD_PRELOAD=$LD_PRELOAD:/lib/x86_64-linux-gnu/libjemalloc.so.2\ bundle\ exec\ ruby\ client.rb
```
richardxia, line 25: This isn't right. [`LD_PRELOAD`](https://man7.org/linux/man-pages/man8/ld.so.8.html) needs to be set to a colon-delimited list of shared object files (`.so`). The Ruby command that you have after the `.so` file might cause issues, since I think it's probably expecting there to literally be a file named (with spaces): `"libjemalloc.so.2 bundle exec ruby server.rb"`.

Could you drop everything after the file name?

richardxia: Also, you don't need the two lines, since I just realized you're setting the same shared library twice.

schroerbrian: Ah shoot. I just copied those lines from the conversation, but I see they were just pass the env into their ruby command. Pushed a fix up here: https://github.com/ShelterTechSF/askdarcel-api/pull/682

PR #683, opened 2023-04-04
Update news articles table

schroerbrian:
Adds URL column to `news_articles` table. Also adds migration to remove defunct `news_article` table. When I ran my migrations locally the `service_at_locations` table showed up in the schema as well. I'm not exactly sure why. Anyone have an idea?

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -353,15 +353,6 @@ ActiveRecord::Schema.define(version: 2023_04_04_230604) do
     t.index ["service_id"], name: "index_schedules_on_service_id"
   end
 
-  create_table "service_at_locations", force: :cascade do |t|
-    t.bigint "service_id"
-    t.bigint "address_id"
-    t.bigint "schedule_id"
-    t.index ["address_id"], name: "index_service_at_locations_on_address_id"
-    t.index ["schedule_id"], name: "index_service_at_locations_on_schedule_id"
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -362,6 +353,15 @@
     t.index ["service_id"], name: "index_schedules_on_service_id"
   end
 
+  create_table "service_at_locations", force: :cascade do |t|
```
schroerbrian, line 356: This popped up when I ran my migration. Anyone know what may be going on here?

richardxia: I'm not exactly sure how Rails represents these things internally, but I wonder if you have a local dev DB that's a bit out of sync with the real DB and that has this table. Could you try reverting the changes to schema.rb, wiping your DB, regenerating your DB, and then running your new migration again to see if it still adds this table?

Another possibility is that our migrations might already be out of sync with schema.rb, and that if you ran all the migrations, you don't get the same result as schema.rb.

Might also be good to check prod and staging to see if this table exists in those environments.

schroerbrian: Awesome thanks for the tip. I confirmed that `service_at_locations` is not in Prod or Staging DB. I reset my DB locally and ran the migration again. It did no re-add the `service_at_locations` table to the schema.

richardxia (approved):
Awesome, LGTM!

PR #684, opened 2023-04-13
Update README.md

jjfreund:


richardxia (approved):
Sorry for the belated review, but this LGTM!

schroerbrian (commented):
I can't remember why we chose to do this. Do we not want to support macOS based development anymore?

richardxia:
> I can't remember why we chose to do this. Do we not want to support macOS based development anymore?

I think it's because it's out of date and we haven't really put the effort into keeping it up to date. We could probably keep if around if someone wanted to actually go through the steps and make sure they're up to date and match the equivalent steps of the Docker-based setup.

lgarofalo (approved):

lgarofalo (approved):

PR #690, opened 2023-05-12
687 authorization schema

schroerbrian:
UPDATE (June 16th)
I've added some updates per our Dev meeting on Wednesday June 14th:
- Permissions are now only assignable to groups. Users can no longer have permissions
- The Permissions table has been revamped. I've added columns for `resource` and `service` foreign keys and a constraint that allows only one of those fields per record. Permissions are dependent on their associated resource or service and will be deleted if their associated object is deleted
- I changed the Permissions "permission" column to "action" (the change request table uses "action" so I just followed that). I've also changed its type from string to enum. Enums are defined in the `Permission.rb` file

I think that's it for now. Thanks all!

******

Update: Should be good to review now, thanks all!

This is my first go at an authorization design for allowing authenticated users to edit certain organizations/resources or services. 

This PR adds Groups and Permissions table. Users can belong to many groups. An individual user or a group can have many permissions. Permissions enable a given user or group to edit a specific object (for our purposes, a service or resource). The permission model specifies the editable objects foreign key.

See comments in models for more details. Any comments on the overall design or the code itself would be much welcome, thanks!


db/schema.rb
```diff
@@ -264,6 +271,19 @@
     t.index ["service_id"], name: "index_notes_on_service_id"
   end
 
+  create_table "permissions", force: :cascade do |t|
+    t.string "description"
+    t.string "permission", null: false
+    t.string "object_type", null: false
+    t.integer "object_pk", null: false
```
richardxia, line 278: It looks like the default column type for primary keys in Rails (with a Postgres backend) is bigint, not, integer, so maybe this should be a bigint to match?

app/models/permission.rb
```diff
@@ -0,0 +1,26 @@
...
+#  object_type :string           not null
+#  object_pk   :integer          not null
+#  user_id     :bigint
+#  group_id    :bigint
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#
+# The permissions system provides a way to assign specific permissions
+# to users and groups of users for specific objects. The object_type
+# is intended to be either a Resource or Service and the object_pk
```
richardxia, line 19: Thanks for writing this comment and for adding the schema in comments! I miss being able to look see the schema at a glance without having to open up schema.rb!

Could we document specifically what `object_type` should look like? Is it literally the strings `"Resource"` and `"Service"`, spelled with a capital initial letter and in singular form? With Rails its a little hard to tell what the canonical name for a table is, since the models are generally spelled in singular form, while the tables are spelled in plural form (and if you use a word with an irregular plural, you end up with weird things like "feedbacks", which isn't technically a word). If they're supposed to be `"Resource"` and `"Service"`, maybe you could put double quotes around them in the comment to more strongly suggest that those are literally what the string value should be?

db/schema.rb
```diff
@@ -264,6 +271,19 @@
     t.index ["service_id"], name: "index_notes_on_service_id"
   end
 
+  create_table "permissions", force: :cascade do |t|
```
richardxia, line 274: Is a Permission supposed to be the representation of an individual permission, or is it supposed to be a "many-to-many" association table? The name would suggest the former, but the fact that you have all the foreign keys on this table implies that it's the latter, where a row in this table can only represent at most one user and/or group at a time.

If this is just the latter, and if the plan isn't to have a DB object represent the former, then maybe we should give this a different name? In Django, I am used to Permissions being the word used to represent, for the lack of a better term, a "permission", the right to perform a certain action or access certain resources, and Permissions have a separate many-to-many join table with both Users and Groups. See https://blog.ovalerio.net/archives/1576 for an entity–relationship diagram.

I'm not sure if this is the best name, but perhaps we could call the many-to-many association something like an Entitlement, meaning whether a particular user or group is _entitled_ to some permission? I've seen the term used in similar contexts, such as in the following articles:

- https://arnon.dk/why-you-should-separate-your-billing-from-entitlement/
- https://developer.apple.com/documentation/bundleresources/entitlements

richardxia (commented):
Great work getting this started! This mostly looks good to me, but I think my main question is about how Permissions are modeled and whether they represent the "right to perform an action" or whether the represent the "association between a user/group and a right to perform an action".

db/schema.rb
schroerbrian: I have updated this to follow the first scenario, that of a representation of an individual permission with a separate many-to-many join tables for both Users and Groups. Thanks for the detailed explanation! I think that following this system is more clear.

db/schema.rb
schroerbrian: Good catch!

jjfreund (changes requested):

app/models/permission.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: permissions
+#
+#  id          :bigint           not null, primary key
+#  description :string
+#  permission  :string           not null
```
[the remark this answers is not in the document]
schroerbrian: Initially my idea was to limit permission values to CRUD operations. Then @richardxia pointed out that in the future we may want to include some custom permissions, depending on the sorts of features that we build out for case workers. I kind of like the idea of leaving it permission values open to allow for more flexibility. That said, I assume most permissions may at least at first be CRUD related.

app/models/user.rb
```diff
@@ -1,4 +1,17 @@
 # frozen_string_literal: true
 
+# == Schema Information
+#
+# Table name: users
+#
+#  id           :integer          not null, primary key
+#  name         :string
+#  is_superuser :boolean          default(FALSE)
```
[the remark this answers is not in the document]
schroerbrian: Here I'm following Django's built in [auth system](https://docs.djangoproject.com/en/4.2/ref/contrib/auth/#django.contrib.auth.models.User.is_superuser), which uses an `is_superuser` column. I suppose we could also just have a `superuser` group, which could function in the same way. I don't feel too strongly one way or the other, though I think a cut and dry boolean is simpler than checking against a group name of "superuser_group", for example

app/models/permission.rb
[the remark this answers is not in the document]
schroerbrian: I've updated the permission table to have an `action` column (following the name of the change_request column) and set it as an enum in the Permission.rb Rails model

app/models/permission.rb
```diff
@@ -0,0 +1,27 @@
...
+#  object_type :string           not null
+#  object_pk   :bigint           not null
```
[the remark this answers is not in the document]
schroerbrian: Great call. Following our discussion in the dev meeting, I've added the resource and service id fields to associate a permission with a service or resource, which will maintain referential integrity.

jjfreund (approved):
lgtm. We may want to add faker data for permissions and groups in a subsequent PR.

richardxia (approved):

PR #694, opened 2023-05-31
1266 Add internal note column to resource and service

schroerbrian:
Internal note column is designed to hold text data about editing the given resource or service.  https://github.com/ShelterTechSF/askdarcel-web/issues/1266


richardxia (approved):
LGTM!

PR #717, opened 2023-06-13
712 add priority rank to category_relationship table. Return categori…

schroerbrian:
…es in order on category#subcategories action

refs https://github.com/ShelterTechSF/askdarcel-api/issues/712


richardxia (approved):
LGTM, thanks for adding that uniqueness constraint to the DB schema!

PR #719, opened 2023-07-19
Enhance texting api

schroerbrian:
This PR addresses two issues:

1.  It adds support to texting information about resources, which was not previously supported by this API. Resources are listed within our search results and receiving their information may be of interest to our user base; this PR enables this
2. Our texting API was not updated to reflect our support for services that have addresses, which are not inherited from their parent resource's primary address. I've added code that first checks if the selected service has an address and, if not, it falls back to the resource address

There is a corresponding front-end PR [here](https://github.com/ShelterTechSF/askdarcel-web/pull/1281); however, this PR should be able to be merged without breaking the current setup

### DIFF
--- a/app/controllers/textings_controller.rb
+++ b/app/controllers/textings_controller.rb
@@ -65,14 +65,14 @@ class TextingsController < ApplicationController
   end
 
   # rubocop:disable Metrics/MethodLength
-  def get_resource_address(resource_address, phone)
-    if resource_address.any?
+  def get_listing_address(address_array, phone)
+    if address_array.any?
       return {
-        address1: resource_address[0].address_1,
-        address2: resource_address[0].address_2,
-        city: resource_address[0].city,
-        state_province: resource_address[0].state_province,
-        postal_code: resource_address[0].postal_code,
+        address1: address_array[0].address_1,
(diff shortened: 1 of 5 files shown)


app/controllers/textings_controller.rb
```diff
@@ -5,24 +5,24 @@
...
-    update_db(recipient_name, phone_number, service_id)
+    update_db
   end
 
   private
 
-  def update_db(recipient_name, phone_number, service_id)
+  def update_db
+    phone_number = texting_params[:phone_number]
+    recipient_name = texting_params[:recipient_name]
```
richardxia, line 25: I feel like it would be better to source this information from the already-processed `text_data` variable than to re-read these from the HTTP parameters. This would avoid throwing out any data sanitization that we're doing, such as what the `#parse_phone_number` method does.

richardxia (approved):
LGTM, aside one minor comment about getting the texting parameters from the already processed data rather than directly from the HTTP parameters!

app/controllers/textings_controller.rb
schroerbrian: Good thinking, I've updated it!

PR #720, opened 2023-07-27
Added API for adding and updating notes and phones outside the change…

jjfreund:
… request paradigm

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -2,17 +2,30 @@
 class NotesController < ApplicationController
   def create
 
-    if params[:service_id]
-      note = service.notes.build
-      note.note = params[:note][:note]
-      service.save!
-    elsif params[:resource_id]
-      note = resource.notes.build
-      note.note = params[:note][:note]
-      resource.save!
+    note = nil
+    note_json = params[:note]
+    if note_json[:service_id]
+      note = Note.create(service_id: note_json[:service_id], note: note_json[:note])
+      note.save!
+    elsif note_json[:resource_id]
+      note = Note.create(resource_id: note_json[:resource_id], note: note_json[:note])
+      note.save!
+    else
+      render plain: "Improper Note Data", status: 400
+      return
     end
-
+ 
     render status: :created, json: note
+  end
+
+  def update
+
+    note = Note.find(params[:id])
+    note.note = params[:note][:note]
+    note.save
+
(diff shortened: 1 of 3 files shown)


schroerbrian (commented):
Looks great. There are some test failures, but I assume you are working on those

app/controllers/notes_controller.rb
```diff
@@ -2,19 +2,32 @@
...
-      service.save!
-    elsif params[:resource_id]
-      note = resource.notes.build
-      note.note = params[:note][:note]
-      resource.save!
+    note = nil
+    note_json = params[:note]
+    if note_json[:service_id]
+      note = Note.create(service_id: note_json[:service_id], note: note_json[:note])
+      note.save!
```
richardxia, line 9: I the `save` is redundant, since `create` will already make a DB request, though you might want `create!` instead of `create` so that it throws an error if the data is invalid.

app/controllers/phones_controller.rb
```diff
@@ -1,6 +1,15 @@
 # frozen_string_literal: true
 
 class PhonesController < ApplicationController
+  def create
+    phone_json = params[:phone]
+
+    phone = Phone.create(resource_id: phone_json[:resource_id], number: phone_json[:number],
```
richardxia, line 7: Should either be `create!` or you should explicitly check that the returned `phone` object was actually saved to the DB. According to the [docs](https://guides.rubyonrails.org/active_record_validations.html#valid-questionmark-and-invalid-questionmark), if you don't opt for `create!`, then you can check the `.valid?` method on the returned object.

config/routes.rb
```diff
@@ -48,6 +48,8 @@
     end
   end
   resources :notes do
+    post :create
+    put :update
```
richardxia, line 52: I think these routes are already handled automatically by the `resources` method, according to https://guides.rubyonrails.org/routing.html#crud-verbs-and-actions. In fact, I think you're actually adding different routes compared to the Rails builtin ones, since this will create the `create` route at `/notes/create`, whereas the builtin is just at `/notes` (with a `POST` method), and yours will create the `update` route at `/notes/update`, whereas the builtin is `/notes/:id/`, where `:id` is the ID of the resource in the URL itself rather than as a parameter.

richardxia (approved):
Changes mostly LGTM, but you're failing some Postman tests.

PR #725, opened 2023-10-12
Validate auth0 token

schroerbrian:
I lifted most of this code from Auth0's Rails tutorial [here](https://auth0.com/docs/quickstart/backend/rails/interactive). This can be merged in as is, given it isn't being used yet to lock down any of our endpoints. The Secured#authorize method can be included in any given controller as a `before_action` (with specific controller methods listed out as needed) to restrict access. Let me know if there's any feedback/requests.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,8 +1,11 @@
 # frozen_string_literal: true
 
 require "jwt"
+require "auth0_client"
 
 class ApplicationController < ActionController::API
+  include Secured
+
   rescue_from ActionController::ParameterMissing do
     head :bad_request
   end
--- a/app/controllers/concerns/secured.rb
+++ b/app/controllers/concerns/secured.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+module Secured
+  extend ActiveSupport::Concern
+
+  REQUIRES_AUTHENTICATION = { message: 'Requires authentication' }.freeze
+  BAD_CREDENTIALS = {
+    message: 'Bad credentials'
+  }.freeze
+  MALFORMED_AUTHORIZATION_HEADER = {
+    error: 'invalid_request',
+    error_description: 'Authorization header value must follow this format: Bearer access-token',
+    message: 'Bad credentials'
+  }.freeze
+
+  def authorize
+    token = token_from_request
+
+    return if performed?
+
+    validation_response = Auth0Client.validate_token(token)
+
+    return unless (error = validation_response.error)
+
+    render json: { message: error.message }, status: error.status
+  end
+
+  private
+
+  def token_from_request
+    authorization_header_elements = request.headers['Authorization']&.split
+
+    render json: REQUIRES_AUTHENTICATION, status: :unauthorized and return unless authorization_header_elements
+
+    render json: MALFORMED_AUTHORIZATION_HEADER, status: :unauthorized and return unless authorization_header_elements.length == 2
+
+    scheme, token = authorization_header_elements
+
+    render json: BAD_CREDENTIALS, status: :unauthorized and return unless scheme.downcase == 'bearer'
+
+    token
+  end
+end
--- a/app/controllers/concerns/secured.rb
+++ b/app/controllers/concerns/secured.rb
@@ -12,6 +12,8 @@ module Secured
   }.freeze
 
   def authorize
+    # This method can be run as a before_action to validate user's token prior to
+    # allowing access to endpoint
     token = token_from_request
 
     return if performed?
--- a/app/controllers/concerns/secured.rb
+++ b/app/controllers/concerns/secured.rb
@@ -4,9 +4,7 @@ module Secured
   extend ActiveSupport::Concern
 
   REQUIRES_AUTHENTICATION = { message: 'Requires authentication' }.freeze
-  BAD_CREDENTIALS = {
-    message: 'Bad credentials'
-  }.freeze
+  BAD_CREDENTIALS = { message: 'Bad credentials'}.freeze
   MALFORMED_AUTHORIZATION_HEADER = {
     error: 'invalid_request',
     error_description: 'Authorization header value must follow this format: Bearer access-token',
--- a/app/controllers/concerns/secured.rb
+++ b/app/controllers/concerns/secured.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# This concern is included in our ApplicationController.
+# It instantiates an Auth0Client, which calls the validate_token method
+# to validate Auth tokens included in requests from the front-end.
 module Secured
   extend ActiveSupport::Concern
 
@@ -11,16 +14,16 @@ module Secured
     message: 'Bad credentials'
   }.freeze
 
+  # The authorize method can be run as a before_action within a controller to validate
+  # a user's token prior to allowing access to an endpoint.
   def authorize
-    # This method can be run as a before_action to validate user's token prior to
-    # allowing access to endpoint
     token = token_from_request
 
     return if performed?
 
     validation_response = Auth0Client.validate_token(token)
-
-    return unless (error = validation_response.error)
+    error = validation_response.error
+    return unless error
 
     render json: { message: error.message }, status: error.status
   end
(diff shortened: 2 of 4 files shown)


lib/auth0_client.rb
```diff
@@ -0,0 +1,50 @@
...
+
+require 'jwt'
+require 'net/http'
+
+class Auth0Client
+  Error = Struct.new(:message, :status)
+  Response = Struct.new(:decoded_token, :error)
+
+  def self.domain_url
+    "https://#{Rails.application.config.x.auth0.domain}/"
```
richardxia, line 11: It might be safer for us to build up these URLs with the Ruby `URI::HTTPS` method: https://ruby-doc.org/stdlib-2.5.1/libdoc/uri/rdoc/URI/HTTP.html#method-c-build (these are the docs for `URI::HTTP`, but I think the API is the same for `HTTPS` but they just don't have good docs for them).

I'm mostly concerned about the trailing slash that you put here, since you're also making the assumption later in the file that the trailing slash exists so that you can join it with the `.well-known/` path. I've just hit lots of issues in the past with two different parts of a code base making incorrect assumptions about whether a trailing slash will exist or not, and these kinds of things all go away if you build up URLs using a more structured approach, since the builders will guarantee that a slash will be added when joining a domain with a path and they usually also prevent double slashes (e.g. `https://example.com//.well-known/jwks.json), which can sometimes mess with string comparisons of URLs that are equivalent but not string-equal.

If it sounds like a reasonable change, do you think you change this so that we can use the URI class more? The docs aren't great, but I think either you can have this method return a `URI` object, and then have the callers append a new path to it if they need to, or you can just have this method only return the domain, without the `https://` or the trailing `/`, and then have the callers construct the `URI` objects by separately passing in the `host:` and `path:` keys.

lib/auth0_client.rb
```diff
@@ -0,0 +1,50 @@
...
+                 algorithm: 'RS256',
+                 iss: domain_url,
+                 verify_iss: true,
+                 aud: Rails.application.config.x.auth0.audience,
+                 verify_aud: true,
+                 jwks: { keys: jwks_hash[:keys] }
+               })
+  end
+
+  def self.jwks
```
richardxia, line 25: I think I'd probably rename this to something that more clearly indicates that it's going to go fetch a file from the internet. `self.jwks` sounds like it's just going to return some hardcoded value, kind of like in instance variable in most languages, but it's doing a lot more than just that.

Maybe `fetch_jwks` or `get_jwks`?

richardxia: Also, this might be more of a personal style preference, but I feel like I'd prefer most of the methods in this class to all be merged together a single one. As a reader, when you see many methods, then you tend to think "oh, what is the API of this method, and who is expected to call it?" But if these aren't actually being used in a lot of places, they could really just be replaced with local variables, where you know that the variables do not get used outside of the method.

lib/auth0_client.rb
```diff
@@ -0,0 +1,50 @@
...
+    jwks_uri = URI("#{domain_url}.well-known/jwks.json")
+    Net::HTTP.get_response jwks_uri
```
richardxia, line 27: Is this something we should be caching? I'm a bit concerned about performance and reliability if every request to one of our auth-guarded routes also has to perform an HTTP request to Auth0 before being able to validate JWTs.

I only did a quick Google search on it, but I found this [StackOverflow post](https://stackoverflow.com/a/52025767) that says that the JWKS keys are usually valid for months at a time. That StackOverflow post also says that the ordinary HTTP Cache Control headers on the HTTP response should let us know how long it should be valid for, so we could even build in a little cache expiration thing that first checks if our JWKS tokens are valid, and only if it isn't, then fetch a new one from Auth0.

richardxia (commented):
I think @lgarofalo has more experience than me on JWTs and stuff, so he might know if there's a general strategy for the caching of JWKS that we should be following.

app/controllers/application_controller.rb
```diff
@@ -1,8 +1,11 @@
 # frozen_string_literal: true
 
 require "jwt"
+require "auth0_client"
 
 class ApplicationController < ActionController::API
+  include Secured
```
waynekoepcke, line 7: I know nothing about Ruby but is this all you have to do to enable SSL?
Don't we need to add a certificate?  As well

richardxia: No, this line mixes in the `Secured` module that he defined in another file in this PR, at app/controllers/concerns/secured.rb. This PR doesn't do anything related to SSL; we terminate SSL at the Kubernetes ingress level, so by the time the request hits the Rails server, it's already speaking HTTP, not HTTPS.

waynekoepcke: ah ok thanks

app/controllers/concerns/secured.rb
```diff
@@ -0,0 +1,43 @@
...
+  def authorize
+    # This method can be run as a before_action to validate user's token prior to
+    # allowing access to endpoint
+    token = token_from_request
+
+    return if performed?
+
+    validation_response = Auth0Client.validate_token(token)
+
+    return unless (error = validation_response.error)
```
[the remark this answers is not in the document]
schroerbrian: Agreed, much more clear. Ruby syntax is funky

app/controllers/concerns/secured.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+module Secured
+  extend ActiveSupport::Concern
```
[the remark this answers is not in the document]
schroerbrian: I've added some basic comments in this file, which I will flesh out as we begin to use this utility. I haven't thought too much about how it will work with Authz checks. We may have to create a separate concern for that, which will draw upon the Roles/Groups tables we created some months ago

lib/auth0_client.rb
schroerbrian: Sure, that trailing slash certainly makes things brittle! I changed this to construct the base URL using the URI:HTTPS library, which I assign to a class constant rather than constructing it within a method. I then append paths where needed using the URI.join method

lib/auth0_client.rb
schroerbrian: Thank you, that is helpful feedback. I've consolidated some of the methods in the class and renamed jwks to the more descriptive "fetch_jwks"

lib/auth0_client.rb
```diff
@@ -0,0 +1,43 @@
...
+    unless jwks_response.is_a? Net::HTTPSuccess
+      error = Error.new(message: 'Unable to verify credentials', status: :internal_server_error)
+      Response.new(nil, error)
+    end
+
+    jwks_hash = JSON.parse(jwks_response.body).deep_symbolize_keys
+    decoded_token = decode_token(token, jwks_hash)
+    Response.new(decoded_token, nil)
+  # rubocop:disable Lint/ShadowedException
+  rescue JWT::DecodeError, JWT::VerificationError
```
richardxia, line 39: I think the Rubocop error is because `JWT::VerificationError` is a subclass of `JWT::DecodeError`, which means that it's redundant to specify `JWT::VerificationError` because `JWT::DecodeError` will already handle it.

Maybe remove the redundant error, which should allow you to remove the `rubocop:disable` comment?

richardxia (approved):
Thanks for making those changes, this LGTM!

For the JWKS caching thing, it's probably something we want to do, but I suppose it doesn't need to block this PR.

lib/auth0_client.rb
schroerbrian: Ahh thanks! That was stumping me

lib/auth0_client.rb
schroerbrian: Great point. I dug a bit into Auth0's docs, as well as community questions, and it seems like JWKS keys issued by this endpoint may not have an expiration date AT ALL. We could cache them indefinitely and/or for some arbitrary time, like a month? I think it is maybe a good idea to punt this for now and maybe it bundle it up with our user authorization implementation. I created a ticket for this in the meantime: https://github.com/ShelterTechSF/askdarcel-api/issues/726 

- https://community.auth0.com/t/jks-validity-period/59407
- https://auth0.com/docs/secure/tokens/json-web-tokens/locate-json-web-key-sets

PR #727, opened 2023-11-30
Adding tables for bookmarks and folders

jjfreund:


db/schema.rb
```diff
@@ -259,6 +269,16 @@
...
 
+  create_table "news_article", force: :cascade do |t|
+    t.string "headline"
+    t.datetime "effective_date"
+    t.string "body"
+    t.integer "priority"
+    t.datetime "expiration_date"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+  end
```
richardxia, line 280: What's up with this? This doesn't look like it's related to the rest of this PR, and it looks like a duplicate of the `news_articles` table created below, but this one is singular, not plural.

db/schema.rb
```diff
@@ -69,12 +69,13 @@
   end
 
   create_table "bookmarks", force: :cascade do |t|
-    t.string "identifier"
-    t.datetime "date_value"
-    t.integer "id_value"
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-    t.index ["identifier"], name: "index_bookmarks_on_identifier", unique: true
+    t.integer "rank"
```
richardxia, line 72: Could we use a different word than "rank"? I think there's a desire to have an actual user ranking/rating system in the future, so this column name could be confusing when that happens. How about "order" or "position"/"pos"?

lgarofalo (approved):

richardxia (approved):

PR #728, opened 2023-12-11
Add missing URL parameter to breaking news article

GeorgeCloud:
Prior to this PR, articles would only receive a URL via a PUT to the NewsArticle Controller.

Related to this [ticket](https://github.com/ShelterTechSF/askdarcel-web/issues/1311)


richardxia (approved):
Thanks for fixing this! If it's not too much trouble, do you think you could try updating the [Postman](https://github.com/ShelterTechSF/askdarcel-api/blob/32c56e0268b6398878e5dad01d05edae165e9db9/postman/AskDarcel%2520API.postman_collection.json#L2661) test to catch this issue, by checking that the API response includes the news article with the URL field set? Let me know if you have any questions about getting the Postman tests running.

richardxia (approved):
Sweet, thanks for adding the test! This LGTM to merge!

Issue #730, opened 2024-01-04
Breaking News Edit Page UI enhancements


richardxia:
When we initially built out the Breaking News feature, we decided to focus our designers' time on the user-facing UI, but not the internal edit view. The current edit page, while functional, has a few strange or potential misleading behaviors, so this is a list of things we can change to improve it:

- It's easy to mistake the "Add Breaking News Article" button for the "Save" button. I think this could be mitigated in two ways:
  - Moving the "Save" and "Delete" buttons to the bottom of each form, above the "Add Breaking News Article" button
  - Changing the text of the "Save" button to "Create" for when we create a new article, vs. leaving it as "Save" when we edit an existing one
- Fix API endpoint so that articles without an effective date are still shown. It's confusing to fill out a news article without an effective date, save it, and then have it disappear when you refresh the edit page
  - Alternatively, make the Effective Date a required field both in the DB and in the UI, so that you can't save if it's not filled in
- Some user visible indication that the Save button has actually performed an action. Right now, hitting the Save button doesn't change the UI at all, so the user is left wondering whether it actually saved. Even an `alert()` call would be fine here.

These are just some things that came to mind, but feel free to edit this ticket with anything else you come up with.

PR #731, opened 2024-02-08
Add ability to query for active articles via NewsArticle model

GeorgeCloud:
These changes are needed for this [Breaking News Edit Page UI enhancements](https://github.com/ShelterTechSF/askdarcel-api/issues/730).


richardxia (approved):
Nice job on this, this LGTM!

FYI, @schroerbrian, @lgarofalo, @jjfreund, the next time we deploy, we'll need to deploy both this and the associated PR, https://github.com/ShelterTechSF/askdarcel-web/pull/1319, at the same time, since this one changes the default behavior of the `/news_articles` index endpoint to return all articles, not just the active one. I think this is the right change to make, as opposed to trying to do it in a backwards-compatible way, but we just have to deploy both at the same time.

richardxia:
BTW, you're failing the lint checks, so you should fix them up. It looks like the failures are autofixable, so if you just run `docker-compose run --rm api bundle exec rubocop -a`, it should automatically fix them.

PR #732, opened 2024-02-09
689 user columns and endpoints

schroerbrian:
This adds name and organization columns to user table, which we will need for saving case worker users. This also creates a UserController with a `create` endpoint as well as a `user_exists` endpoint. The f/e checks the `user_exists` prior to making requests to create a new user in both Auth0 and our DB. This PR also adds tests for said endpoints.


katerina-kossler:
can you add a description to your PR?

schroerbrian:
> can you add a description to your PR?

Sure, I have just added one

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+class UsersController < ApplicationController
+  def create
+    user = User.create({ email: params[:email], name: params[:name], organization: params[:organization] })
+    render status: :created, json: UserPresenter.present(user)
+  end
+
+  def user_exists
+    user = User.find_by_email(params[:email])
+    render json: { user_exists: user ? true : false }
+  end
```
richardxia, line 12: We actually shouldn't expose an unprotected endpoint like this, since this is a security/privacy vulnerability. The security best practice is to not reveal anything about a user unless that user is authenticated. This even applies to things like registering an account and resetting passwords. See https://github.com/OWASP/CheatSheetSeries/blob/90ed5b2acc0516cf7f0e87a4029fba54b9020ade/cheatsheets/Authentication_Cheat_Sheet.md#authentication-and-error-messages

I haven't fully read through your frontend PR yet, but if possible, if this is only called in contexts where the user has already authenticated against Auth0, could we change this so that:

1. It's only accessible to authenticated users, and
2. It can only return results for the same email address that was authenticated by Auth0? I don't know what data is stored in the auth token that Auth0 will send back to the app, but we can only use the data within that token in order to look up whether a User record has been created in our app, since we can only trust that the data contained within the token is authenticated, not external parameters like the `email` param you have here.

    If Auth0 gives us an opaque identifier rather than an email address, then we'll want to store that ID in our database instead of (or maybe in addition to) the user's email.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class UsersController < ApplicationController
+  def create
```
richardxia, line 4: Thinking about the security of the app, I think we should also be protecting this endpoint. Unlike my comment about `user_exists`, the risk isn't about leaking data, but instead it's about allowing an unauthorized person write data about another user that they shouldn't be able to.

With the auth token that Auth0 sends us, we should validate that the user described by the token indeed has control of the email being passed in as the address here.

richardxia (commented):
I have a few security concerns about the changes here. I think due to its central role in our user identity system, the UsersController probably needs authentication and authorization logic on all of its controller methods. I've forgotten the details about exactly what Auth0 sends us, but if you have any questions and have that info handy, I'm more than happy to talk more about the security concerns and how we can safely deal with this.

schroerbrian:
> I have a few security concerns about the changes here. I think due to its central role in our user identity system, the UsersController probably needs authentication and authorization logic on all of its controller methods. I've forgotten the details about exactly what Auth0 sends us, but if you have any questions and have that info handy, I'm more than happy to talk more about the security concerns and how we can safely deal with this.

Thanks for reviewing, Richard! I think I can rejigger things to lock this down. I am going to lay out the current order of operations just to clarify it in my own mind. 

1. User submits sign-up form
2. We hit the userExists endpoint to check if user exists in our DB
3. If not, we create them in Auth0
4. User verifies email on our site, we forward to Auth0
6. Auth0 creates user. NB: If they already exist, Auth0 will simply auth them against their servers
7. Following successful Auth0 creation, we hit the create endpoint
8. We establish the user session in React/Front-end

Here is my proposed new order to lock it down:

1. User submits sign-up form
2. We create them in Auth0
3. User receives verification code in their email
4. User verifies email on our site, we forward to Auth0
5. Auth0 creates user. NB: If they already exist, Auth0 will simply auth them against their servers
6. Regardless, we will postpone establishing the user session on the front-end with the data passed back from Auth0 until we've created the user in our DB
7. Hit the create endpoint; check auth0 token to ensure user email matches the auth0 email; 
8. if user exists, send error back advising user to sign in instead
9. Upon successful user creation, establish user session

This is a very rough draft and may change slightly but i'll start tweaking it now. thanks again!

app/controllers/users_controller.rb
schroerbrian: I have made some updates here. Now this endpoint requires that the request include an Auth0 token, which is received when the f/e successfully auths against Auth0, and the user's designated auth0 user ID. Our rails app validates this token by hitting Auth0's servers; a success response includes the user ID for which the token has been assigned. the #create method compares this user ID against the one received from the frontend to ensure that they match.

One question is do we want to be storing the Auth0 userId in our database as well to handle cases where the user changes their email?

I think we may have discussed caching the `authorize` before_action, which I can work on doing in a future PR.

richardxia: > One question is do we want to be storing the Auth0 userId in our database as well to handle cases where the user changes their email?

Yes, I think we should store that, and that we should use it as the primary way to associate an Auth0 JWT to a User object in our DB. Thanks for bringing that up.

> Lastly, I think we may have discussed caching the `authorize` before_action, which I can work on doing in a future PR.

That sounds good to me, and it could even possibly be a ticket for a new volunteer, since I think it would be relatively narrowly scoped, and it's not on the critical path (even if the performance would be fairly bad without it).

app/controllers/concerns/secured.rb
```diff
@@ -22,8 +22,13 @@ def authorize
     return if performed?
 
     validation_response = Auth0Client.validate_token(token)
+
     error = validation_response.error
-    return unless error
+    unless error
+      # Create user_id instance variable so that it can be compared against userId recieved from client
+      @user_id = validation_response.decoded_token[0]['sub']
```
richardxia, line 29: Minor suggestion: Since this is the Auth0 ID, not the primary key of the `users` table in our DB, how about we name the variable `@user_external_id`? (I debated calling it `@user_auth0_id`, but wanted something less specific to the auth provider we're using.)

richardxia (approved):
LGTM!

app/controllers/concerns/secured.rb
schroerbrian: That is a great idea. I have used that nomenclature for the User table column as well.

app/controllers/users_controller.rb
schroerbrian: I went to create a caching ticket and then saw that it was already created, haha.  I've been on this project for too long (or not long enough?) https://github.com/orgs/ShelterTechSF/projects/10/views/1?pane=issue&itemId=42087115

PR #736, opened 2024-02-28
Ability to query for active articles via NewsArticle model

GeorgeCloud:
These changes are needed for this [Breaking News Edit Page UI enhancements](https://github.com/ShelterTechSF/askdarcel-api/issues/730).

Messed up the last PR. Sorry about that!


richardxia (approved):
No worries, this LGTM!

BTW, for future reference, it should be possible to fix up the old PR without having to create a new one. If you have everything correct in your local branch, then all you need to do is run `git push` with the `--force` flag to push that up to GitHub, overwriting the history of what was previously up on that branch on GitHub. It's not critical, but it's usually nice to keep using the same PR, since it keeps all the discussion in one place.

PR #744, opened 2024-06-11
Update categories and eligibilities for DCYF.

richardxia:
This adds a migration that makes a large number of changes to categories and eligibilities, to match the changes requested by DCYF. Each group of categories and elgibilities is handled a little bit differently, but they roughly fall into one of the following patterns:

- Create a new category/eligibility
- Rename a category/eligibility
- Mass migrate a number of resources and services from a set of old categories/eligibilities to a new one
- Sometimes delete the old categories/eligibilities when doing said mass migration, though in some cases preserve the old ones

This was implemented as a pure data migration that does not change the schema. Since data migrations must be agnostic to the actual models, we cannot use any of the model code, and instead we construct a number of raw SQL queries. These queries have been factored into independent methods so that the main migration code can be read top to bottom with high-level method calls describing the exact changes being made.

This is supposed to follow the instructions in https://docs.google.com/spreadsheets/d/1rCvcITtUdmjOz3IzjNClbgY4UO53xGV8ix7oeeXDtKc/edit#gid=688234094, and I'd appreciate it if someone could double check my work. I've been testing locally on a dump of the production database, and I'm planning to spend some time later tonight to sanity check all my queries.

### DIFF
--- a/db/migrate/20240610165140_update_dcyf_categories_and_eligibilties.rb
+++ b/db/migrate/20240610165140_update_dcyf_categories_and_eligibilties.rb
@@ -38,14 +38,13 @@ class UpdateDcyfCategoriesAndEligibilties < ActiveRecord::Migration[6.1]
       # Academic Support
       create_category 'Academic Support'
 
-      # TODO: Need to get the answer to two issues:
-      # 1) Summer School is not an actual category name
-      # 2) The spreadsheet says to _not_ delete Summer School, unlike all the
-      # others
-      # ['Academic', 'Education', 'Educational Supports', 'Tutoring', 'School Care', 'Summer School'].each do |from|
-      #   migrate_resources_and_services_to_new_category from: from, to: 'Academic Support'
-      #   delete_category from
-      # end
+      ['Academic', 'Education', 'Educational Supports', 'Tutoring', 'School Care'].each do |from|
+        migrate_resources_and_services_to_new_category from: from, to: 'Academic Support'
+        delete_category from
+      end
+      # Summer Programs is special in that we do not want to delete it, only
+      # copy its resources and services over to Academic Support
+      migrate_resources_and_services_to_new_category from: 'Summer Programs', to: 'Academic Support'
 
       # Alternative Education & GED
       create_category 'Alternative Education & GED'
@@ -143,6 +142,8 @@ class UpdateDcyfCategoriesAndEligibilties < ActiveRecord::Migration[6.1]
 
       # Experiencing Homelessness
       rename_eligibility from: 'I am someone experiencing homelessness', to: 'Experiencing Homelessness'
+      migrate_services_to_new_eligibility from: 'Homeless', to: 'Experiencing Homelessness'
+      delete_eligibility 'Homeless'
 
       # Boys
       create_eligibility 'Boys'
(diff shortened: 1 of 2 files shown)


richardxia:
One bug I discovered by manually examining the DB after applying this migration is that this will result in duplicate entries in the many-to-many join tables, since we don't have any uniqueness constraints on them. For example, because `categories_services` doesn't have a unique key on `category_id` and `service_id`, it's possible to insert a second row with the same category and service.

I need to alter the logic to avoid creating duplicates when possible.

richardxia:
One other oddity I noticed is that the data in Algolia will be stale after running this migration, so we'll probably want to retrigger a mass reindexing after applying this migration.

schroerbrian (approved):
Wow this is great - really well-thought out how you managed the changes and check for a category’s existence or non-existence first. I don’t currently have my personal computer with me but this code LGTM. Thanks!

schroerbrian:
> One other oddity I noticed is that the data in Algolia will be stale after running this migration, so we'll probably want to retrigger a mass reindexing after applying this migration.

Good thought. I suppose this maybe be since these records are created outside of the Resource/Service.rb models, which is where the indexing code lives

db/migrate/20240610165140_update_dcyf_categories_and_eligibilties.rb
```diff
@@ -0,0 +1,287 @@
...
+      rename_category from: 'Housing', to: 'Housing & Rental Assistance'
+
+      ['Housing Assistance', 'Emergency Rental Assistance'].each do |from|
+        migrate_resources_and_services_to_new_category from: from, to: 'Housing & Rental Assistance'
+      end
+
+      # Immigration Assistance
+      create_category 'Immigration Assistance'
+
+      ['Citizenship & Immigration'].each do |from|
```
katerina-kossler, line 86: Why is this in an array, is this meant to be two categories? or waiting to hear back on adding others?

katerina-kossler (approved):
lgtm! had one smalll q but seemed like this would all do what we want!

db/migrate/20240610165140_update_dcyf_categories_and_eligibilties.rb
richardxia: Ah, yeah, good call. This is just one category; I misunderstood the original instructions and thought that "Citizenship" and "Immigration" were two different categories, but they're actually just a single one (it's way clearer in code, since string boundaries are clearly marked with quotes). I can remove the loop and just directly call the migration function on this one category.

richardxia: Changed in https://github.com/ShelterTechSF/askdarcel-api/pull/744/commits/c47c404b620c052375967ee32a39f2c41297b56f

db/migrate/20240610165140_update_dcyf_categories_and_eligibilties.rb
```diff
@@ -0,0 +1,285 @@
...
+
+      # Playgroups
+      create_category 'Playgroups'
+
+      # Academic Support
+      create_category 'Academic Support'
+
+      # TODO: Need to get the answer to two issues:
+      # 1) Summer School is not an actual category name
+      # 2) The spreadsheet says to _not_ delete Summer School, unlike all the
```
richardxia, line 43: Got confirmation that:

1. "Summer School" should actually be "Summer Programs", and
2. "Summer Programs" indeed should _not_ be deleted, unlike all the other ones. I'll update this code and make sure that we specifically do not delete "Summer Programs"

db/migrate/20240610165140_update_dcyf_categories_and_eligibilties.rb
```diff
@@ -0,0 +1,285 @@
...
+      rename_eligibility from: 'High School Students', to: 'High School'
+
+      # College
+      rename_eligibility from: 'College Students', to: 'College'
+
+      # CSF
+      create_eligibility 'CCSF'
+
+      # Experiencing Homelessness
+      rename_eligibility from: 'I am someone experiencing homelessness', to: 'Experiencing Homelessness'
```
richardxia, line 145: While getting the answers to the previous question, I also learned that we need to migrate all the services and organizations from the eligibility named `Homeless` into this new `Experiencing Homelessness` tag.

db/migrate/20240610165140_update_dcyf_categories_and_eligibilties.rb
richardxia: Fixed in https://github.com/ShelterTechSF/askdarcel-api/pull/744/commits/30e74a9ffe6cc6fd996be1e0ca675a52fcb40e6c

db/migrate/20240610165140_update_dcyf_categories_and_eligibilties.rb
richardxia: Fixed in https://github.com/ShelterTechSF/askdarcel-api/pull/744/commits/30e74a9ffe6cc6fd996be1e0ca675a52fcb40e6c

richardxia:
I'm going to do one last set of sanity checking the data resulting from this migration with my local DB, specifically focusing on the last set of changes I pushed up, but otherwise I'm planning to merge this later tonight and maybe doing a deploy tomorrow.

richardxia:
Just finished spot-checking the data, and I think the new changes I made are working correctly. I'll leave this open for a few more hours, in case if @lgarofalo wanted to check it out, but I'll merge it regardless by the end of tonight.

PR #748, opened 2024-08-05
Add `boosted_category_id` column and add to Algolia index.

richardxia:
This adds a `boosted_category_id` column to Services, which is a nullable foreign key to the Categories table. This is intended to be used by the frontend code through Algolia as a possible target of the optional filters feature of Algolia, allowing us to boost search rankings for specific services on specific pages.

This column is added to the Algolia index as `boosted_category`, which holds the string name of the category, and we also add it to the `attributesForFaceting` list so that it can be used for optional filters.

Finally, we change the Algolia index ranking order from its default by placing `geo` after `filters`. This is needed in order for optional filters to have higher precedence than geolocation.

### DIFF
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -9,6 +9,7 @@ class Service < ActiveRecord::Base
 
   belongs_to :resource, required: true, touch: true
   belongs_to :program
+  belongs_to :boosted_category, class_name: 'Category'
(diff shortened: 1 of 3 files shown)


schroerbrian (approved):
pretty cool, thanks for getting this done!

PR #750, opened 2024-08-22
Fix boosted category facet in Resource model.

richardxia:
We recently added a boosted category column on the services table, with the intent of syncing it to Algolia so that we can perform special types of search ranking based on it on the frontend. However, the way we added it resulted in an obscure bug where the feature appeared to work until someone visits the edit page and saves an edit to a resource or service.

The issue is that when we set the `attributesForFaceting` property, we only added the `boosted_category` field to the Service model, not the Resource model. Since we're doing something unusual by sharing the same index for both the Service and Resource model, what actually happens is that the Resource model would sometimes override the `attributesForFaceting` in the Algolia index with its own list, causing that attribute to not be usable in filtering anymore.

This fixes the issue by adding the `boosted_category` field to Resource's list of facets. To prevent this kind of issue from happening in the future, we refactor the list of attributes for faceting into a single constant that both the Service and Resource models share.

### DIFF
--- a/app/models/resource.rb
+++ b/app/models/resource.rb
@@ -44,7 +44,7 @@ class Resource < ActiveRecord::Base
                   index_name: "#{Rails.configuration.x.algolia.index_prefix}_services_search",
                   id: :algolia_id do
       # specify the list of attributes available for faceting
-      attributesForFaceting %i[categories open_times eligibilities associated_sites type]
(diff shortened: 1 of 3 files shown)


lgarofalo (approved):

schroerbrian (approved):
Thanks for fixing this!

Issue #127, opened 2017-01-19
Add verification date to resources and services


jjfreund:
This will allow us to mark resources and services as verified to be correct even in the absence of changes, and display that verification date in the app.

PR #753, opened 2024-10-17
Add name to bookmarks table

katerina-kossler:
part of: [#127](https://github.com/ShelterTechSF/sheltertech-go/issues/127)

changes: 
- new field on bookmarks table (nullable) for backwards compat

notes:
will follow in future to set bookmark name to service name by default (and to any existing bookmarks)

questions:
not sure how we deploy db changes this year


richardxia (approved):
Sorry for the super late review, but this LGTM! Thanks for working on it!

PR #755, opened 2025-02-27
Rename categories to change Covid- prefix to sfsg-.

richardxia:
These were originally named during the beginning of the COVID-19 pandemic, but since then, they've taken on a more general purpose that is specific to the SFSG site, as opposed to the other white-labeled sites.

Running this migration also needs to be followed up with reindexing the Algolia, and this will also require some coordinated changes on the frontend app.


jjfreund (approved):
lgtm

PR #764, opened 2026-01-29
Fix import staging script

schroerbrian:
The import staging script was choking on the bookmarks table due to it having a column named "order", which is a reserved keyword in pgsql. This PR enhances our staging_importer#fast_insert method by quoting table name and column name and thus preventing this from happening with other reserved words.


richardxia (approved):
LGTM! Thanks for fixing this!

schroerbrian:
> LGTM! Thanks for fixing this!

Thanks for the review!

PR #765, opened 2026-02-19
SFSG Tile Updates

dfloo:
Add new migration for category updates

### DIFF
--- a/db/migrate/20260219165140_update_sfsg_tiles.rb
+++ b/db/migrate/20260219165140_update_sfsg_tiles.rb
@@ -20,6 +20,8 @@ class UpdateSfsgTiles < ActiveRecord::Migration[6.1]
       ## sfsg-shelter
       create_category 'Adult Shelter Reservation'
 
+      create_category_service_relationship 'Adult Shelter Reservation', 'Adult Shelter Reservation System'
+
       create_category_relationship 'sfsg-shelter', 'Adult Shelter Reservation'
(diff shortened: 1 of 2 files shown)


db/migrate/20260219165140_update_sfsg_tiles.rb
```diff
@@ -0,0 +1,157 @@
...
+      create_category_relationship 'sfsg-health', 'Hospice' 
+      create_category_relationship 'sfsg-health', 'In-Home Support' 
+      create_category_relationship 'sfsg-health', 'Assisted Living' 
+      create_category_relationship 'sfsg-health', 'Disease Screening' 
+
+      delete_category_relationship 'sfsg-health', 'Coronavirus-Related Urgent Care' 
+      delete_category_relationship 'sfsg-health', 'Coronavirus (COVID-19) Testing' 
+         
+      ## sfsg-shelter
+      create_category 'Adult Shelter Reservation'
```
richardxia, line 21: Do we need to create this as a real category? If we could make this a frontend-only thing, then it wouldn't pollute the actual DB and edit interface with this category.

richardxia (approved):
This LGTM aside from my one question about creating the Adult Shelter Reservation category. Did you run the migration locally on a copy of the prod DB? It's pretty important to do that to make sure that we won't hit any issues when trying to deploy the changes.

richardxia (approved):
New change LGTM!

